Issue #37, opened 2015-01-18
Rotating axis text not working


mswastik:
Rotating axis text is not working with animint
```
fg=ggplot() + geom_point(aes(x=State, y=Prop.Inv, showSelected=Year), data=ss) + theme_animint(width=600, height=400) + theme(axis.text.x=element_text(angle=70, hjust=1))
sg=ggplot() + stat_summary(data=ss, aes(Year, Year, clickSelects=Year), fun.y=length, geom="bar")
map=list(fg=fg, sg=sg)
animint2dir(map ,out.dir="animint")
``` 
doesn't rotate the text on x axis.

While if I simply plot ```fg``` the text in x-axis is rotated but when combine plots with ```animint2dir``` the text in x-axis is horizontal.

tdhock:
Hi, thanks for the bug report. Indeed axis tick text rotation is one of the many ggplot theme options that has not yet been implemented in animint. However I think it should be relatively easy to implement and I have already started work on the rotate branch, so please add any further comments there #39.

PR #39, opened 2015-01-19
Rotated axis labels

tdhock:
#37 reported that axis tick rotation does not work.

This is a first try to translate R/ggplot

```r
theme(axis.text.x=element_text(angle=70, hjust=1)) 
```

to HTML/SVG

```html
<text style="text-anchor: end;" transform="rotate(-70)">
```

So far I have just written renderer support for xangle and xanchor axis properties, so now we need to make the compiler output these values.

Also we need tests and compiler/renderer code for y axis angle rotation.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: animint
 Maintainer: Toby Dylan Hocking <tdhock5@gmail.com>
 Author: Toby Dylan Hocking, Susan VanderPlas, Carson Sievert
-Version: 2014.12.4
+Version: 2015.01.19
 License: GPL-3
 Title: Interactive animations
 Description: An interactive animation can be defined using a list of
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: animint
 Maintainer: Toby Dylan Hocking <tdhock5@gmail.com>
 Author: Toby Dylan Hocking, Susan VanderPlas, Carson Sievert
-Version: 2015.01.19
+Version: 2015.01.27
 License: GPL-3
 Title: Interactive animations
 Description: An interactive animation can be defined using a list of
--- a/NEWS
+++ b/NEWS
@@ -56,9 +56,6 @@ https://github.com/tdhock/animint-examples/blob/master/examples/scaffolds.R
 took 15 minutes to compile on my computer.
 
 OPTIMIZATION: to save disk space, we can compress the TSV file
-database to .tsv.gz files. But for this to work with
-d3.tsv("path/to/file.tsv.gz"), we need a web server that serves .gz
-files with the Content-Encoding: gzip header -- is this possible with
 standard web servers? With servers like gist.github.com, all .json
 files are served with Content-Encoding: gzip, so they use
 http://en.wikipedia.org/wiki/HTTP_compression over the network. In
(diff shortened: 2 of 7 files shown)


tdhock:
Rotated x axis tick text should be working now. Next I will work on implementing the same functions for the y axis.

@mswastik can you please download the code and check if it works for your purposes?

```r
devtools::install_github("tdhock/animint@rotate")
```

mswastik:
Thanks for adding the feature @tdhock the texts are rotated now...but as you can see in the image texts are cutoff due to limited space available for them and x-axis label name did not shifted downward. I think some final adjustments are needed.
image

tdhock:
Thanks for the revealing screenshot! I guess we will need to add some JavaScript code that adjusts the axis margin size in function of the tick text size. @cpsievert implemented some related code with facet margin sizes recently, do you have any ideas about how to do that?

cpsievert:
[Here](https://github.com/cpsievert/animint/blob/master/inst/htmljs/animint.js#L137-142) is the snippet where we measure the amount of padding to add for x/y tick labels. The problem seems to be due to the fact that `measureText()` assumes that the text is at an angle of 0, 90, 180, ...

I'm not immediately sure how to modify `measureText()` to support this feature. Any ideas @srvanderplas?

srvanderplas:
Trig? If you specify the angle, you should be able to get a relatively
decent measure of height and width offset from the plot. Obviously there
will be some nesting issues with the width, but I think it should be fairly
easy to deal with that by only considering the edge cases.

cpsievert:
I'd like to avoid doing unnecessary math, if possible :P

The [getBBox](http://stackoverflow.com/questions/1636842/svg-text-element-width?rq=1) method looks promising, but we'd have to measure SVG elements as opposed to HTML. This should also have the added benefit of avoiding creation of hidden elements.

cpsievert:
@tdhock would you like me to take a stab at re-implementing `measureText()` to use `getBBox()`? I'm thinking it'd be nice to have measureText() take on a css selector to select (existing text) elements and return {height: num, width: num}

srvanderplas:
Sounds like that investment might also pay off if we wanted to fix things
ggplot2 doesn't do that well (expanding axes to accommodate text labels,
etc.)

tdhock:
Yes please go ahead and try to re-implement measureText, and push commits to rotate branch.

tdhock:
And I guess we should add a test for axes labels not overlapping tick labels, but that seems complicated.

cpsievert:
The following code gives me [this](http://bl.ocks.org/cpsievert/062c9c578519cc1f48a6) now.

```s
library(animint)
mtcars$vs <- paste("Some really long label text", mtcars$vs)

p <- qplot(data = mtcars, vs, wt, geom = "point") +
  facet_wrap(~am) +
  theme(axis.text.x = element_text(angle=70, hjust=1)) 

animint2gist(list(plot = p))
```

Let me know if any ideas for more tests

tdhock:
I used Selenium's getLocation and getSize methods to implement a test to make sure tick text does not overlap x axis title.

http://selenium.googlecode.com/git/docs/api/java/org/openqa/selenium/WebElement.html#getLocation%28%29

cpsievert:
@tdhock I like the tests you have here. You have my approval to merge. Please update DESCRIPTION/NEWS accordingly :)

PR #40, opened 2015-01-27
Test

cpsievert:
This is currently broken, but a step closer to an ideal testing framework.

Most importantly, I want an argument to specify which browser to use for testing.

It might also be nice to eventually export a function like `animint::run_test()` one day, so to run a test on brand new code, we'd just have to `devtools::load_all(); run_test()`.

PR #41, opened 2015-01-30
Improve testing infrastructure

cpsievert:


tdhock:
Ok so this branch is for trying to get animint to work with the new servr, right?

and the other branch is for actually making our tests runnable using any browser?

cpsievert:
Yep. I'll probably use the first approach I described [here](https://github.com/yihui/servr/issues/13) on #40 just to get it working and return to this later.

tdhock:
Great, thanks for the reference to that servr issue.

PR #43, opened 2015-03-12
implement test-hjust-text-anchor.R

caijun:
Using test-hjust-text-anchor.R to test Animint, the testing results are as follows

1. Failure(@test-hjust-text-anchor.R#96): geom_text(hjust=0) => <text style="text-anchor: start"> ----------------
style.value does not match 'start'. Actual value: "middle"

2. Failure(@test-hjust-text-anchor.R#103): geom_text(hjust=1) => <text style="text-anchor: end"> -----------------
style.value does not match 'end'. Actual value: "middle"

### DIFF
--- a/tests/testthat/test-hjust-text-anchor.R
+++ b/tests/testthat/test-hjust-text-anchor.R
@@ -47,9 +47,6 @@ objective <- ldply(objective$iteration, function(i) {
 objective2 <- subset(objective, iteration == iteration2)
 
 grad.desc.viz <- function(hjust) {
-  if (!hjust %in% c(0, 0.5, 1)) {
-    warning("unsupported value of hjust was provided.")
-  }
   objective2$hjust <- hjust
   
   contour.plot <- ggplot() + 
@@ -78,7 +75,6 @@ grad.desc.viz <- function(hjust) {
   viz <- list(contour = contour.plot, objective = objective.plot, 
               time = list(variable = "iteration2", ms = 2000), 
               title = "Demonstration of Gradient Descent Algorithm")
-  info <- animint2HTML(viz)
 }
 
 getStyleValue <- function(html, xpath, style.name) {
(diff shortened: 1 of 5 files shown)


tests/testthat/test-hjust-text-anchor.R
```diff
@@ -0,0 +1,115 @@
...
+getStyleValue <- function(html, xpath, style.name) {
+  nodes <- getNodeSet(html, xpath)
+  node.style <- xmlAttrs(nodes[[1]])["style"]
+  pattern <-paste0("(?<name>\\S+?)", ": *", "(?<value>.+?)", ";")
+  style.matrices <- str_match_all_perl(node.style, pattern)
+  style.value <- style.matrices[[1]][style.name, "value"]
+}
+
+test_that('geom_text(hjust=0) => <text style="text-anchor: start">', {
+  info <- grad.desc.viz(hjust = 0)
```
tdhock: in tests it is best to be as explicit as possible, so I would recommend that grad.desc.viz returns an animint list of ggplots and options, and then you take animint2HTML outside of the grad.desc.viz function, e.g.

```r
viz <- grad.desc.viz(hjust=0)
info <- animint2HTML(viz)
```

tests/testthat/test-hjust-text-anchor.R
```diff
@@ -0,0 +1,115 @@
...
+objective <- dat$objective
+objective <- ldply(objective$iteration, function(i) {
+  df <- subset(objective, iteration <= i)
+  cbind(df, iteration2 = i)
+})
+objective2 <- subset(objective, iteration == iteration2)
+
+grad.desc.viz <- function(hjust) {
+  if (!hjust %in% c(0, 0.5, 1)) {
+    warning("unsupported value of hjust was provided.")
```
tdhock: the warning should be added to the animint2dir compiler, not to the grad.desc.viz helper function.

caijun:
I have implemented `get_text_anchor` function in animint.js. 

https://github.com/caijun/animint/blob/master/inst/htmljs/animint.js#L717-L733

Tests of `geom_text(hjust=0) => <text style="text-anchor: start">`, `geom_text(hjust=1) => <text style="text-anchor: end">`, and `geom_text(hjust=0.5) => <text style="text-anchor: middle">` tests in test_hjust_text_anchor.R have successfully been passed. 

Yes, it's easier to understand the code if putting `animint2HTML` outside of the `grad.desc.viz` function. I  have updated test-hjust-text-anchor.R.

As for the `geom_text(hjust=0.5) => <text style="text-anchor: middle">` test, in my grad.desc.viz,  `hjust` aesthetic is given by a variable name rather than a value.
```
geom_text(aes(hjust = hjust)
```

The invalid input of hjust should be checked in `get_text_anchor` function and give a warning. But I don't know how pass warning from Javascript to R.

tdhock:
These changes look good. Although it is possible to check for invalid hjust values at rendering time (and maybe report it using [console.log](http://www.w3schools.com/js/js_output.asp)), I think it makes more sense to check at compile time and then assume we have valid values at rendering time.

Actually there is already R code that should stop the compiler with an error

https://github.com/tdhock/animint/blob/master/R/animint.R#L174-L188

so please change the test to check for an error (not a warning), as we do with hjust for axis labels

https://github.com/tdhock/animint/blob/master/tests/testthat/test-rotate.R#L90-L97

caijun:
I check for invalid hjust values inside the `saveLayer` function of animint compiler.

https://github.com/caijun/animint/blob/master/R/animint.R#L322-L330

The test for invalid hjust values has been changed to check for an error.

https://github.com/caijun/animint/blob/master/tests/testthat/test-hjust-text-anchor.R#L114

All tests of test-hjust-text-anchor.R have been successfully passed, and you can merge this branch.

tdhock:
Looks good to me. Now just update the package version to 2015.03.23 in DESCRIPTION, and add some info about the bug that you fixed in NEWS. After that I will merge into master.

PR #45, opened 2015-03-17
Updated .travis.yml to port travis support to native

da-ta-vinci21:
Since Travis-CI now provides native support for R, this commit updates the <b>.travis.yml</b> file to port it from r-travis to native Travis support for R.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,39 +4,19 @@
 # see the wiki:
 #   https://github.com/craigcitro/r-travis/wiki
 
-language: c
+language: r
 
-before_install:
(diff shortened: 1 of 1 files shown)


cpsievert:
Cool, thanks, looks good to me :+1: 

Go ahead and merge if you'd like @tdhock

cpsievert:
@tdhock hmm, it seems like that would be tedious to maintain...

I would rather that we rely on CRAN versions wherever possible on the master branch. I'm pretty sure the problems we experienced earlier were because we were on the bleeding edge...

tdhock:
ok, let's keep depending on the non-specific versions for now (and cross our fingers that the other guys do not introduce backwards-incompatible changes).

PR #48, opened 2015-03-25
aes(linetype) for tall and wide rects

tdhock:
it was simple to add support but we should also add tests


tdhock:
I also noticed a bug in facet strip labels, when there is only 1 row or column of facet_grid.

we should add this degenerate case to the facet tests @cpsievert

tdhock:
I added some tests for the trivial case of 1 row and/or 1 column of facets, in test-facet-trivial.R

The last example (plot named Sk) fails the test that the x axis label should be visible. @cpsievert can you please take a look at that when you get a chance?

tests/testthat/test-widerect.R
```diff
@@ -67,14 +70,32 @@ wb.facets <-
...
+      style.list[[node.i]] <- a.vec[["style"]]
+      a.vec <- xmlAttrs(node)
+      sizes <- as.numeric(a.vec[c("height", "width")])
+      expect_true(all(sizes > 0))
+    }
+    style.vec <- do.call(c, style.list)
+    dash.mat <- str_match_perl(style.vec, dasharrayPattern)
+    dash.table <- table(dash.mat[, "value"])
+    ## There should be 2 unique values of stoke-dasharray.
+    expect_equal(length(dash.table), 2)
```
tdhock: This new test makes sure that the aes(linetype=status) in the geom_tallrect and geom_widerect gets translated to 2 different values of stroke-dasharray in the <rect> elements.

tests/testthat/test-facet-trivial.R
```diff
@@ -0,0 +1,50 @@
...
+  expect_that(trans.y, is_less_than(400))
+}
+
+test_that("facet_grid(1 row and/or 1 column) is fine", {
+  info <- animint2HTML(viz)
+  expect_axes("kk", 1, 1)
+  expect_axes("kx", 1, 1)
+  expect_axes("xk", 1, 1)
+  expect_axes("kS", 3, 1)
+  expect_axes("Sk", 1, 3)
```
tdhock: this last test does not pass since the xtitle's y position is too large to be rendered (it is below the 400 x 400 rect of the SVG). I guess this is a problem with the facet computations. @cpsievert can you try to fix those facet computations when you get a chance?

xtitle-below-svg

tdhock:
some tests were giving errors, but now they are fixed.

now we just need to fix the facet computations.

tdhock:
on a somewhat related note, I realized that R colors defined in parameters such as `geom_line(colour="grey50")` were not getting correctly translated to `rgb(127, 127, 127)` in JSON.

This is now fixed, and there are now tests for the compiler and renderer.

cpsievert:
Nice work, looks good! :shipit:

PR #51, opened 2015-03-31
Namespaces

cpsievert:


cpsievert:
`R CMD check` now produces only two notes of concern (to CRAN, at least). If we ever want to submit to CRAN, we'd have to explain/argue/plead why we need to do these things:

(1) [Unexported objects imported by ':::' calls:](https://travis-ci.org/tdhock/animint/builds/56610059#L1599)
(2) [no visible binding for global variable](https://travis-ci.org/tdhock/animint/builds/56610059#L1609)

Let me know what you think @tdhock

tdhock:
to avoid that NOTE for tests_init, we could move it to testthat/functions.R

the only way I know of to avoid the ::: problem is by copy/pasting the relevant code from the other pkg into animint

tdhock:
anyway I don't really care about submitting to CRAN do you?

tests/testthat/test-gist.R
```diff
@@ -0,0 +1,7 @@
+context("gists")
+
+test_that("animint2gist() returns an object of class 'gist'", {
+  g <- animint2gist(list(p = qplot(1:10)), screenshot = FALSE)
+  expect_is(g, "gist")
+  gistr::delete(g)
```
tdhock: so travis posts a gist and then deletes it with every check now?

If so, AWESOME!

tdhock:
looks good to me +1 merge when you like

tests/testthat/test-gist.R
cpsievert: Believe it! 

excited_beyond_belief

cpsievert:
If we move `test_init()`, `test_run()`, etc. to testthat/functions.R, then we can't export them, which makes testing more painful. 

It might be nice to release to CRAN someday to increase exposure, but maintaining a CRAN package can a great source of unnecessary pain. I like Yihui's rant [here](https://stat.ethz.ch/pipermail/r-devel/2013-August/067283.html) about dealing with these sorts of issues (I think the `utils::globalVariables` bit he refers to is a hack for (1) above)

PR #57, opened 2015-04-30
use cran packages

cpsievert:


cpsievert:
@tdhock  I forgot there are some problems with gistr 0.1, but I was just informed 0.2 will hit CRAN hopefully tomorrow, so let's wait for that...

cpsievert:
I just ported some `animint2dir()` fixes to #62 since the rest of these commits are crap.

PR #58, opened 2015-05-01
try wercker

cpsievert:


cpsievert:
Testing this out on [my fork](https://github.com/cpsievert/animint) first since I need to be owner to create the application

PR #62, opened 2015-05-04
Get both travis & wercker builds working

cpsievert:
See discussion [here](https://github.com/travis-ci/travis-ci/issues/3849)


cpsievert:
@tdhock here are some notes on using wercker to inspect builds locally

https://github.com/tdhock/animint/wiki/Testing#inspect-test-builds-locally-with-wercker

I just used it identify an [issue with RSelenium](https://github.com/ropensci/RSelenium/pull/46). Worked great!

tdhock:
since you merged the "test global variables" commits here, I propose to just merge this branch and close #64

cpsievert:
@tdhock we should probably fix [this test error](https://travis-ci.org/tdhock/animint/builds/61537071#L1642) before merging. Any idea why that's occurring? Looks like that error was occurring over on #64 as well

tdhock:
```
> remDr$getCurrentUrl()
[[1]]
[1] "http://localhost:4848/animint-htmltest/"
```

cpsievert:
Looks like there are two globals: "plot" and "animint" since we [do this](https://github.com/tdhock/animint/blob/7a0d63ab6c9e426094a2833407ec32441d668035/inst/htmljs/index.html#L17-L19) in index.html

@tdhock do you think that's OK to have "plot" as a global?

cpsievert:
I suppose hard coding "animint-htmltest" isn't a big deal since it's hard coded in a couple other places...

tdhock:
that is fine to hard code animint-htmltest I think.

it is of course fine to have "plot" as a global, since it is defined in index.html (the test is to make sure that there are no global variables being introduced in animint.js)

however "plot" does not show up in diff.vars on my local tests with firefox. for portability of this test across firefox and phantomjs I suggest we filter "plot" out of diff.vars, and then `expect_identical(diff.vars, "animint")`

tdhock:
thats great that travis tests are passing again!

cpsievert:
Well, they *were* passing because I had the shiny tests commented out :disappointed: 

On the bright side, tests are passing for me locally on Firefox, but I'm getting the same stalling on phantomjs as wercker during the "rotate" tests ([seems like a lot of people run into this](https://github.com/ariya/phantomjs/issues/11526)).

tdhock:
we could use `testthat::skip_on_travis` https://github.com/hadley/testthat/blob/master/man/skip.Rd for the shiny tests?

cpsievert:
I'm not exactly sure why, but phantomjs didn't like [this business](https://github.com/tdhock/animint/blob/7a0d63ab6c9e426094a2833407ec32441d668035/tests/testthat/functions.R#L181-L182) in the rotate tests (at least on wercker and my machine), but 56d4294 should fix that problem.

I have a feeling that calling those two methods (one right after another) on the same element was the problem.

I'll probably end up using something like `testthat::skip_on_travis()` for the shiny tests

cpsievert:
@tdhock I'm sort of at a loss on what to do here. The remaining problem has to do with [this code](https://github.com/tdhock/animint/blob/c09f0591c0f6cb1b8728c8ca8815af6ceb5b12fa/tests/testthat/test-widerect.R#L117-L120), which times out on both my machine and wercker (it works fine for me locally with Firefox though). I tried replacing those methods with pure JS, but this doesn't quite work...any ideas/suggestions?

```r
remDr$executeScript('return document.getElementById("updates_ms").value = 3000')
```

tdhock:
I guess the reason why it doesn't work with pure JS is because the R `e$sendKeysToElement(list("3000", key="enter"))` code activates the `.on("change"` function, which updates `Animation.ms`. The test is supposed to make sure that editing the text box results in changing the `Animation.ms` variable.

Can we just skip this test on phantomjs?

cpsievert:
@tdhock ok, I think this is ready. I also updated the testing portion of the wiki. Will merge once I have your +1

tdhock:
Thanks for this great work Carson. Tests pass on my local computer using firefox. +1 please merge when you like.

PR #64, opened 2015-05-06
test global variables

tdhock:
This branch introduces a test to make sure that the animint.js file introduces only 1 thing (animint) into the global JavaScript environment.

Thanks to @johndharrison for explaining how to implement this https://github.com/ropensci/RSelenium/issues/47

Currently there are 4 objects that animint.js defines:

```
> diff.vars
[1] "animint"                "measureText"            "linetypesize2dasharray"
[4] "isArray"               
```


tdhock:
The test passes now since I moved the 3 helper functions inside of the animint function.

PR #65, opened 2015-05-13
Knit .Rmd file to create tornado plots

kferris10:
The .Rmd file had not been knitted recently.  As a result, the animint plots in some of the subdirectories had not been created so the link did not work.


cpsievert:
Awesome, thanks! Would you mind changing the width (not the height) of [this map to 970](https://github.com/kferris10/animint/blob/946931eb77230d390bdb8b91f7ecae268b90e713/index.Rmd#tornado-example) before I merge?

cpsievert:
Also, not sure why, but this plot isn't showing up 

http://kferris10.github.io/animint/tornado-ts-bar/index.html

tdhock:
Hey @kferris10 thanks for your first PR!

Another suggestion for the tutorial would be to have "list of other options" link to this wiki page which explains some of the non-standard options possible in animint https://github.com/tdhock/animint/wiki/Advanced-features-present-animint-but-not-in-ggplot2

kferris10:
Thanks for the suggestions guys.  I think I've [got them updated](http://kferris10.github.io/animint/).  What is the best way to update this PR?  Should I just create a new one?

cpsievert:
Yep, looks great, thanks!

PR #68, opened 2015-05-18
Pirates example

kferris10:
Create a visualization of the pirates data and remove it from the TODO lists.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: animint
 Maintainer: Toby Dylan Hocking <tdhock5@gmail.com>
 Author: Toby Dylan Hocking, Susan VanderPlas, Carson Sievert
-Version: 2015.06.16
+Version: 2015.06.22
 License: GPL-3
 Title: Interactive animations
 Description: An interactive animation can be defined using a list of
--- a/NEWS
+++ b/NEWS
@@ -197,6 +197,8 @@ does not respond to mouse clicks while the data is loading.
 DSL: Implement shape aesthetic with d3.svg.symbol()
 https://github.com/mbostock/d3/wiki/SVG-Shapes#symbol_type
 
+RENDER: Determine why border of tiles in pirates example does not seem to be rendering.
+
 2015.05.29
 
 BUGFIX: warnings for opening pid.txt, closing unused connection and generating make_text.Rd.
--- a/NEWS
+++ b/NEWS
@@ -199,11 +199,13 @@ https://github.com/mbostock/d3/wiki/SVG-Shapes#symbol_type
 
 RENDER: Determine why border of tiles in pirates example does not seem to be rendering.
 
+2015.06.22 
+
+EXAMPLE: Incorporate example with pirates data.
+
 2015.06.16
 
 RENDER: Implement panel backgrounds, panel borders, and grid lines in plots.
-
-EXAMPLE: Incorporate example with pirates data.
 
 2015.05.29
 
--- a/NEWS
+++ b/NEWS
@@ -156,8 +156,6 @@ DSL: how to specify the layout of SVG elements on the page?
   means that we should layout the plots in a <table> with
   2 rows and 1 column.
 
-EXAMPLE: pirates data: world map?
-
 EXAMPLE: US state population since 1790, compare w/ other countries?
 http://en.wikipedia.org/wiki/List_of_U.S._states_by_historical_population
 http://fr.wikipedia.org/wiki/Histoire_démographique_de_la_France
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ We keep a TODO list at top of the [NEWS](https://github.com/tdhock/animint/blob/
 - BUG: things which used to work but have stopped working. Tests should be added to prevent these. For example at one point in 2013, animint rendered the correct number of legend entries for the WorldBank viz, but in the beginning of 2014 animint rendered too many legend entries. 
 - GGPLOT: things which ggplot2 supports but animint does not yet support. For example facets, coord_equal.
 - DSL: changes to the animint domain-specific language (DSL) which would allow interactive/animated features. These involve changes to how we define the ggplots, and how the compiler works. For example custom alpha/color/etc for selected geoms, using selected.alpha/selected.color/etc aesthetics.
-- EXAMPLE: examples to show off animint features, which should affect neither the compiler nor renderer. For example the pirates data set.
+- EXAMPLE: examples to show off animint features, which should affect neither the compiler nor renderer. 
 - RENDER: changes to the JavaScript rendering code which would result in better interactive plots, without having to change the definition of the ggplots. For example rendering a selection widget for every selection variable.
 - OUTPUT: different output formats for viewing/sharing an interactive animation.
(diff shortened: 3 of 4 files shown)


cpsievert:
I wouldn't worry about the build failure. I think for the gist test to work properly, you'll need write access to this repo, so maybe @tdhock should grant you those permissions...

Did you look into using `geom_density2d()` as opposed to manually computing those bins and using `geom_tile()`?

kferris10:
I did briefly, but I like the tiles better.

image

cpsievert:
Oops, I meant `geom_bin2d()`, but I forgot that animint doesn't support it :(

```r
S <- matrix(c(1, .5, .5, 1), nrow = 2)
obs <- mvtnorm::rmvnorm(1000, sigma = S)
df <- setNames(data.frame(obs), c("var1", "var2"))
p <- ggplot(df, aes(var1, var2)) + geom_bin2d() +
  scale_fill_gradient(trans = "log")
animint2dir(list(p = p))
```

```r
 Error in saveLayer(L, df, meta) : 
  bin2d is not supported in animint. Try using geom_tile() and binning the data yourself.
```

This looks good to me, @tdhock go ahead and merge it you'd like

:+1:

srvanderplas:
geom_density2d can be used with fill, if that helps. I agree about the
tiles, though - they seem to make a nicer visualization.

kferris10:
I think the tiles look better too.  Thanks for your help!

tdhock:
About the blue background, using a blue geom_rect is a clever workaround with the current code, but I would suggest this is a good time to actually implement that `theme` feature in animint. Basically that would mean (1) telling the compiler to extract the blue color from the theme element in `parsePlot`  https://github.com/tdhock/animint/blob/master/R/animint.R#L61 and then (2) tell the renderer to color the `<svg>` element, inside of `add_plot` https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L180

About the viz itself, I would suggest to add a second selection variable. For example, `geom_tile(aes(clickSelects=location))` in the second map, and then change the time series plot to `geom_line(aes(showSelected=location))` and use `selector.types=list(location=multiple)` ... this reminds me of a similar viz I did with some climate change data, see http://cbio.ensmp.fr/~thocking/animint-climate-change/index.html and source https://github.com/tdhock/climate-change/blob/master/figure-stations.R for possible inspiration.

tdhock:
About computing your own `geom_tile` versus having ggplot2 compute `geom_density2d` etc, `geom_tile` is preferable if you want to use clickSelects in the tiles.

tdhock:
I added @kferris10 as a collaborator with push access to tdhock/animint.

kferris10:
Does that mean that I should just push the repository to tdhock/animint?  If so, to the master branch?

tdhock:
in general you should try to avoid pushing to master if at all possible.

the better way to do it is to create a PR and then wait until you get comments and a +1 from either Carson or myself.

for this PR you already have Carson's +1, but you may want to address some of my comments, if you want to try to fix theme background colors, or try a re-design of the pirates viz with more interaction variables.

But it is fine if you don't want to do those extra things in this PR. In that case just push (to your kferris10:pirates_example branch) with a version update to NEWS (include a link to this PR) and DESCRIPTION, then click Merge pull request above.

kferris10:
I think I'll take a stab at your suggestions today and tomorrow.  I'm going to close this PR and I'll open a new one when those updates are ready.

tdhock:
if you end up pushing code to your kferris10:pirates_example branch then it is fine to re-open this same PR.

kferris10:
I was planning to pushing it to tdhock:pirates_example to hopefully advoid any permission issues with the Travis tests.  Does that make sense?

cpsievert:
Since your just adding an example here, I wouldn't worry about the tests. Do whatever is easiest for you.

kferris10:
Okay, I'll just open this one back up.

@tdhock I'll work on your first suggestion (fixing the panel.background) and submit it as a separate PR.  For your second suggestion, I've updated [the visualization](http://bl.ocks.org/kferris10/raw/138c81b780fc13627548/).  I like your idea.  What do you think?

tdhock:
Looks good! A few more suggestions:

Can you increase the default animation time delay? Personally I think it is
too fast and I interactively increased it to 1 second.

For the first map, can you color the dots by type of attack or some other
variable?

For the second couple of linked plots, the reader/user needs some way to
decode which lines correspond to which selected squares. One way to do that
would be to add a geom_text(aes(label=id, showSelected=id)) to both plots.

kferris10:
[Updated link](http://bl.ocks.org/kferris10/raw/c68296a1e7b025611850/)

1.  I kind of like the faster animation.  I think that it makes it a little more interesting to view.

2. I don't think there's enough information available in the data to do much with the descriptions.  I haven't looked at them closely, but I think they are too specific to do much with.

3. Good call.  I also added clickSelects to the lines.

tdhock:
ok, your call for the animation speed, no problem.

about the geom_text id labels, what you have is a good start. I think it
would be good to also try adding a black geom_text with the same id to the
map.

and I would add a geom_tallrect behind the second time series lines, or
maybe combine the two time series plots into a single facetted plot with
facet_grid(variable ~ ., scales="free") -- which would result in aligned x
axes (dates).

kferris10:
[Updated link](http://bl.ocks.org/kferris10/raw/70d5ac0b1689d15c8abe/)

Thanks for your suggestions. I thought about adding the ids to the map, but I think it gets a bit noisy.  Especially if you select some tiles near each other, it gets hard to read pretty quickly.

tdhock:
Thanks for your efforts Kevin. In terms of design this version looks best so far to me, even though the ids on the map can get a little busy. I think it is worth it to be able to quickly see the correspondence between the lines and the highlighed black squares.

Finally it seems like this viz reveals a bug in the animint.js renderer. After having already selected a few black squares on the second map, selecting a new year sometimes briefly moves the black square to an incorrect position. Do you see that too? or is it just on my web browser?

tdhock:
the moving black squares is just a problem for me on firefox, not on chrome. so I guess we should not worry about this.

kferris10:
Great! I won't have Wifi til Friday so I'll update this PR then

kferris10:
[Updated visualization](http://bl.ocks.org/kferris10/raw/0b87a424fb2b2acc8bbf/)

I've gone back to the original code which uses `theme(panel.background = ...)` so this requires #78 to display the lightblue backgrounds.

kferris10:
Is it okay if I go ahead and merge this example?

tdhock:
Running your example code gave me

```
thocking@silene:~/R/animint-kferris10(pirates_example)$ Rscript inst/examples/pirates.R 
Loading required package: methods
Loading required package: proto

Attaching package: ‘dplyr’

The following object is masked from ‘package:stats’:

    filter

The following objects are masked from ‘package:base’:

    intersect, setdiff, setequal, union

Error in nrow(.) : object '.' not found
Calls: %>% ... mutate_ -> mutate_.tbl_df -> mutate_impl -> .Call -> nrow
Execution halted
```

kferris10:
Hmmm...it works for me.  Is it possible that you are using an older version of `dplyr` or `magrittr`?

tdhock:
You were right, I was in dependency hell! 

I finally got it to work with

```r
works_with_R("3.2.0",
             ggplot2="1.0.1",
             dplyr="0.4.2",
             tidyr="0.2.0.9000",
             maps="2.3.6",
             lubridate="1.3.3",
             stringr="0.6.2")
```

+1

PR #70, opened 2015-05-18
Get tests working on windows

cpsievert:


cpsievert:
@kferris10 let me know if you get this working on Windows and then I'll merge

kferris10:
It doesn't stop with an error anymore!

The only thing that happens is I get a bunch of warnings saying that 

    Warning messages:
    1: In file(con, "r") :
      cannot open file 'C:/Users/chessa015/Documents/packages/animint/tests/testthat/pids.txt': No such file or directory
    2: closing unused connection 5 (C:/Users/chessa015/Documents/packages/animint/tests/testthat/pids.txt) 

`tests_init("firefox")` also works.  The browser opens and I don't get an error saying failed to connect anymore!

cpsievert:
Great! Have you been able to do `test_run()` successfully?

cpsievert:
I suppose we could _really_ make sure tests are working on Windows with this, but that might be overkill... https://github.com/krlmlr/r-appveyor

kferris10:
I think it's working.  The testing starts up, and most of the tests work fine

    > tests_run()
    aesthetics : List of 1
     $ showSelected: symbol foo
    .
    animation : ...
    One character value : ......

But when I get down to `geom_widerect`

    geom_widerect : 8...............

    1. Failure(@test-facet-lines.R#62): a <path> is rendered for every country ------
    length(node.set) not equal to nrow(countries)
    Mean relative difference: 1

    2. Failure(@test-interactive-facets.R#91): red lines are only drawn in panel 2 --
    length(PANEL2lines) not equal to 6
    Mean relative difference: 1

    3. Failure(@test-interactivity.R#172): default is 2 <path> and <text> elements --
    length(nodes) not equal to 2
    Mean relative difference: 1

    4. Failure(@test-interactivity.R#174): default is 2 <path> and <text> elements --
    length(nodes) not equal to 2
    Mean relative difference: 1

    5. Failure(@test-tooltip.R#65): aes(clickSelects) means show 'variable value' ---
    tooltips does not match 'country'. Actual value: ""

    6. Failure(@test-tooltip.R#72): aes(tooltip) means show tooltip -----------------
    tooltips does not match 'not NA'. Actual value: ""

    7. Failure(@test-tooltip.R#78): aes() means show no tooltip ---------------------
    length(rect.nodes) not equal to 1
    Mean relative difference: 1

    8. Failure(@test-widerect.R#74): widerect renders a <rect> for every year -------
    length(node.set) not equal to nrow(years)
    Mean relative difference: 1

     Error: Test failures 

Does this mean that `tests_run()` is working and that there's just a problem in `geom_widerect()` somewhere?

cpsievert:
Bummer, looks like widerect is the _eighth_ error :(

Is this happening with phantomjs or firefox?

I know you mentioned you _could_ run a linux box, so that end up being the better option...

cpsievert:
Actually, if you're up to the task, it'd be great if you could identify _why_ those errors are occurring (and I'm happy to offer advice for debugging tests, but it'll be hard to offer much more help if I can't access your machine)

cc @tdhock

tdhock:
here are a couple of ideas which may or may not help resolve your testing problems:

can you manually execute the code in the tests that are failing using firefox? and then visually check to see if the plots are rendering correctly?

or if the remote-controlled browser is not working, try substituting `animint2dir` for `animint2HTML` then and manually/visually check?

this may help determine if the problem is the browser you are using or if it is the RSelenium remote controller.

more generally perhaps we should ask ourselves if it is worth the effort to support testing on windows? On the one hand I think our linux-based testing infrastructure is quite adequate and I would rather spend development time on implementing new features. On the other hand, if it is not too much effort it would be nice to know that animint works on windows.

kferris10:
So this is interesting, when running `tests_init()` I don't get the same errors every time.  I've gotten as few as 8 and as many as 20.  However, when running `tests_init("firefox")`, I've only gotten one error.

    1. Failure(@test-shiny.R#37): animint plot renders in an interactive document ---
    length(circles) >= 1 isn't true

I figured I'd post this and see what you guys thought before looking into it more closely myself.

Minor note, `tests_init()` works if I am not connected to Wifi, but `tests_init("firefox")` does not

    > tests_init("firefox")
    Undefined error in RCurl call.
    Starting selenium binary. To shut it down, run: 
    remDr$closeWindow() 
    remDr$closeServer()
    No such file or directoryfailed to load external entity "http://selenium-release.storage.googleapis.com"

cpsievert:
That's great news that tests are basically working on Firefox! 

(I've had some problems with that shiny test too, but it passes on Travis and Wercker, so :see_no_evil:). 

BTW, what version of phantomjs are you using?

@tdhock I think we should merge this. Is that cool with you?

tdhock:
+1 go ahead and merge it when you like.

PR #72, opened 2015-05-21
bugfix for transparent fill, many layers

tdhock:
#48 fixed the bug that converts color parameters such as `geom_line(color="grey50")` to the proper RGB values, but this also introduced a bug which converts "transparent" to white.

This PR fixes that bug. It also adds tests, including one that makes sure that `toRGB(NA)=="transparent"` as in ggplot2.

### DIFF
--- a/R/animint.R
+++ b/R/animint.R
@@ -568,12 +568,14 @@ saveLayer <- function(l, d, meta){
           chunk.vec <- g.data[[chunk.var]]
           counts <- table(chunk.vec)
           if(length(counts) == 1){
-            stop("only 1 chunk") # do we ever get here?
-          }
-          if(all(counts == 1)){
-            FALSE #each chunk has only 1 row -- chunks are too small.
-          }else{
+            ##stop("only 1 chunk") # do we ever get here?
             TRUE
+          }else{
+            if(all(counts == 1)){
+              FALSE #each chunk has only 1 row -- chunks are too small.
+            }else{
+              TRUE
+            }
           }
         }else{
           FALSE
--- a/R/animint.R
+++ b/R/animint.R
@@ -1112,7 +1112,7 @@ getLegend <- function(mb){
   if(length(dataframes)>0) {
     data <- merge_recurse(dataframes)
   } else return(NULL)
-  data <- lapply(nrow(data):1, function(i) as.list(data[i,]))
+  data <- lapply(1:nrow(data), function(i) as.list(data[i,]))
   if(guidetype=="none"){
     NULL
   } else{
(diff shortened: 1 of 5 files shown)


R/animint.R
```diff
@@ -568,12 +568,14 @@ saveLayer <- function(l, d, meta){
           chunk.vec <- g.data[[chunk.var]]
           counts <- table(chunk.vec)
           if(length(counts) == 1){
-            stop("only 1 chunk") # do we ever get here?
-          }
-          if(all(counts == 1)){
-            FALSE #each chunk has only 1 row -- chunks are too small.
-          }else{
+            ##stop("only 1 chunk") # do we ever get here?
             TRUE
```
tdhock: this `stop` is not necessary (I added it while debugging the initial implementation of the compiler code which divides each data set into chunks). It was never being run in the past because most the data sets of most showSelected layers have more than one value of the showSelected variable. However I am currently developing such a data viz, so we need to get rid of this `stop`.

R/animint.R
```diff
@@ -856,7 +858,9 @@ animint2dir <- function(plot.list, out.dir = tempfile(),
         update.vars <- L$mapping[is.ss | is.cs]
         has.var <- update.vars %in% names(L$data)
         if(!all(has.var)){
-          print(update.vars[!has.var])
+          print(L)
+          print(list(problem.aes=update.vars[!has.var],
+                     data.variables=names(L$data)))
           stop("data does not have interactive variables")
```
tdhock: this error message is now a bit more user-friendly: `print(L)` should help the user identify which layer caused the error in a complex multi-layer plot.

R/animint.R
```diff
@@ -568,12 +568,14 @@ saveLayer <- function(l, d, meta){
...
-            stop("only 1 chunk") # do we ever get here?
-          }
-          if(all(counts == 1)){
-            FALSE #each chunk has only 1 row -- chunks are too small.
-          }else{
+            ##stop("only 1 chunk") # do we ever get here?
             TRUE
+          }else{
+            if(all(counts == 1)){
+              FALSE #each chunk has only 1 row -- chunks are too small.
```
cpsievert: `else if () {}` could get rid of an extra set of parenthesis...

R/animint.R
cpsievert: I like to avoid `print()` statements since they are hard to suppress, but I think I'm OK with this (since it's essentially part of an error message).

R/animint.R
tdhock: ok, I made that change.

tdhock:
I also added a test to make sure the renderer respects both numeric and character linetypes, and legend ordering via `scale(limits)`

this code works for me, can I get a +1 before I merge with master?

cpsievert:
Assuming the tests pass, this has my :+1:

PR #74, opened 2015-05-21
Getting started with interactive legends

kferris10:
This is just a rough start to implementing interactive legends.  @cpsievert thought it would be good to discuss where exactly we want this to go.  He has a couple ideas which are slightly over my head so I'll let him explain.

I've [put together a working example](http://kferris10.github.io/animint_dev/) so you can see the general idea.  I also explain the changes I've made there.  I'm pretty sure that this is the only example for which this will work. 

I'd also love any suggestions for tests that might be useful here.  Thanks!


cpsievert:
This is a wonderful start @kferris10! Could you also provide the R code used to create this example?

kferris10:
Absolutely.  [Here you go](https://github.com/kferris10/animint_dev/blob/gh-pages/iris_plots.R)

tdhock:
That is very interesting! Can you get it to work with multiple selection? I would expect all you have to do is specify the option list `selector.types=list(Species="multiple")`.

kferris10:
Looks like it looks fine with multiple selection.  [Link](http://kferris10.github.io/animint_dev/)

tdhock:
Great! Two ideas: 

1. can you get the legend text to fade out (50% alpha transparency) when it is not selected?

2. interesting that you manually specified `aes(color=Species, clickSelects=Species)` for one plot and `aes(color=Species, showSelected=Species)` for the other. I would suggest also trying to modify the animint2dir compiler so that it works for a single plot with just `aes(color=Species)` -- in that case the behavior that I would expect is that the compiler automatically/internally adds the showSelectedColor aesthetic, assigns a default first selection of `first=list(Species=c("versicolor", "virginica", "setosa"))`, and sets a default of `selector.types=list(Species="multiple")`. So initially the plot would render all 3 Species, but then clicking the legend entries would show/hide each one.

kferris10:
These both seem like they'll be a little bit tricky to me.  Right now, this works by just calling update_selector() when the user clicks on a row.  So I'd have to modify the compiler so that a clickSelects attribute is added to each of these geoms (I think).  Does that make sense?

tdhock:
indeed that is a good start to add a call to update_selector in the
renderer animint.js code for each legend entry.

about modifying the compiler I suggest you put the code inside of
parsePlot. I think the call to getLegendList
https://github.com/tdhock/animint/blob/master/R/animint.R#L142 should be
moved up near the beginning of the function. Then check the output of
getLegendList and edit the appropriate layers before the calls to saveLayer
https://github.com/tdhock/animint/blob/master/R/animint.R#L36

I don't think it makes sense to add a clickSelects aesthetic to each layer
that makes a legend. For example in the WorldBank viz
http://sugiyama-www.cs.titech.ac.jp/~toby/animint/WorldBank/viz.html the
geom_line has aes(color=region) and aes(clickSelects=country) so it would
only make sense to add aes(showSelectedColor=region). Does that make sense?
or am I missing some other use case involving another data set/viz?

kferris10:
Sorry to take so long to get back to this.  I got sidetracked with the panel stuff.

I'm a little bit confused about what you're thinking here.  In the "Great! Two ideas:" comment, it seems like you're suggesting we make it so that any time `aes(color=Species)` code is entered, legend interactivity is automatically implemented.  But in the next comment ("indeed that is a good start..."), it sounds like you're suggesting we consider implementing another aesthetic: `aes(showSelectedColor=region)`.  

Also, I just want to run this by everybody.  They way I'm thinking about it, there would only be clickSelects/showSelected interactivity for legends with discrete values.  For example in [this plot](http://bl.ocks.org/kferris10/raw/2192837f1f0d10f4c44b/), users could select groups by clicking on the Species legends but clicking on the Sepal.Width legends wouldn't do anything.  Does that approach make sense or is there some example that I'm missing?

tdhock:
Sorry for the unclear comments.

Yes I was suggesting that we make it so that any time aes(color=Species) code is entered, legend interactivity is automatically implemented.

the `aes(showSelectedColor=region)` was an idea for how to implement this.

First I should have explained the fact that the animint.R compiler uses any aesthetic that contains "showSelected" is interpreted as a selector variable. For example, in a geom with `aes(showSelected=var1, showSelected2=var2, showSelectedFoo=var3)` all three will be interpreted as selection variables.

The designer (person who writes the ggplot code) will write only `aes(color=region)` but the animint.R compiler will automatically add a showSelectedColor=region aesthetic. The compiler will later interpret that to treat region as a selector (showSelected) variable.

Finally, yes I agree that this only makes sense for legends with discrete values.

kferris10:
Okay, I think I understand what you're saying.  With this new push, you are o longer required to use `aes(clickSelects = Species)`.  You do still have to include a `showSelected` statement however.  I'll try and remove this requirement next.  

I've [added an example](https://github.com/tdhock/animint/blob/interactive_legends/inst/examples/interactive-legends.R) to the list of examples so that you guys can see where things currently are.  I don't think this is super helpful for people who are new to animint so I'll delete it once we are done with this development.  A [live version](http://bl.ocks.org/kferris10/raw/a62a990434195a4f48c8/) is here.  You can click on any of the `species` legends and the selected points will update!

The cool thing is that this is no longer limited to the iris example!  You should be able to use it with any dataset if you're interested.

R/animint.R
```diff
@@ -22,11 +22,34 @@ parsePlot <- function(meta){
...
     L <- meta$plot$layers[[layer.i]]
+    
+    ## If any legends are specified, add showSelected aesthetic
+    for(legend.i in seq_along(plot.meta$legend)) {
+      var_name <- plot.meta$legend[[legend.i]]$vars
+      var <- L$data[[var_name]]
+      if(is.factor(var) | is.character(var)) {
+        L$mapping$showSelectedcolour <- as.symbol(var_name)
+      }
+    }
```
kferris10: This is my initial idea for adding showSelected aesthetics.  @tdhock is this what you had in mind?

kferris10:
[Updated link](http://bl.ocks.org/kferris10/raw/31ef63b9b261664b5ec4/)

I still need to update things a little bit, but it's a start.

R/animint.R
```diff
@@ -22,11 +22,40 @@ parsePlot <- function(meta){
...
+      # the actual values for that variable
+      var <- L$data[[var_name]]
+      # checking if it is a discrete variable
+      if(is.factor(var) | is.character(var)) {
+        # if it is, adding a showSelected aesthetic for it
+        L$mapping$showSelectedcolour <- as.symbol(var_name)
+      }
+    }
+    ## need to call ggplot_build again because I've added to the plot
+    meta$built <- ggplot2::ggplot_build(meta$plot)
```
kferris10: This is my initial idea for adding showSelected aesthetics. @tdhock is this along the lines of what you had in mind?

tdhock: Very nice work Kevin! This is a great, almost exactly what I had in mind.

R/animint.R
```diff
@@ -22,11 +22,54 @@ parsePlot <- function(meta){
...
+    for(legend.i in seq_along(plot.meta$legend)) {
+      # the type of legend: colour, fill, etc
+      legend_type <- plot.meta$legend[[legend.i]]$legend_type
+      # the name of the variable used in this legend
+      var_name <- plot.meta$legend[[legend.i]]$vars
+      # the actual values for that variable
+      var <- L$data[[var_name]]
+      # checking if it is a discrete variable
+      if(plyr::is.discrete(var)) {
+        # adding a showSelected aesthetic for each legend entry
```
kferris10: @tdhock I think this takes care of the "aes name a function of the legend aes name" issue.  The only issue now is that instead of `showSelectedFill` it does `showSelectedfill`.  This way makes more sense to me (since the aesthetics are lowercase in ggplot2).

R/animint.R
```diff
@@ -22,11 +22,54 @@ parsePlot <- function(meta){
...
+        for(i in legend_type) {
+          temp_name <- paste0("showSelected", i)
+          L$mapping[[temp_name]] <- as.symbol(var_name)
+#           browser()
+        }
+        # if first is not specified, add all to first
```
kferris10: This ensures that if `first` is not specified, the default is to set everything to selected

R/animint.R
```diff
@@ -22,11 +22,54 @@ parsePlot <- function(meta){
...
+        if(is.null(meta$first[[var_name]])) {
+          meta$first[[var_name]] <- unique(var)
+        }
+        # if selectortypes is not specified, set it to multiple
```
kferris10: And this ensures that the default is to use multiple selection

R/animint.R
```diff
@@ -22,11 +22,54 @@ parsePlot <- function(meta){
...
+        # adding a showSelected aesthetic for each legend entry
+        for(i in legend_type) {
+          temp_name <- paste0("showSelected", i)
+          L$mapping[[temp_name]] <- as.symbol(var_name)
+#           browser()
+        }
+        # if first is not specified, add all to first
+        if(is.null(meta$first[[var_name]])) {
+          meta$first[[var_name]] <- unique(var)
+        }
```
kferris10: This ensures that if `first` is not specified, the default is to set everything to selected

R/animint.R
```diff
@@ -22,11 +22,54 @@ parsePlot <- function(meta){
...
+        # if selectortypes is not specified, set it to multiple
+        if(is.null(meta$selector.types[[var_name]])) {
+          meta$selector.types[[var_name]] <- "multiple"
+        }
```
kferris10: And this ensures that the default is to use multiple selection

R/animint.R
```diff
@@ -22,11 +22,54 @@ parsePlot <- function(meta){
...
+      # the actual values for that variable
+      var <- L$data[[var_name]]
+      # checking if it is a discrete variable
+      if(plyr::is.discrete(var)) {
+        # adding a showSelected aesthetic for each legend entry
+        for(i in legend_type) {
+          temp_name <- paste0("showSelected", i)
+          L$mapping[[temp_name]] <- as.symbol(var_name)
+#           browser()
+        }
```
kferris10: @tdhock I think this takes care of the "aes name a function of the legend aes name" issue. The only issue now is that instead of showSelectedFill it does showSelectedfill. This way makes more sense to me (since the aesthetics are lowercase in ggplot2).

kferris10:
As for calling `ggplot_build` twice, I'm sure that there is a way around it, but I'm not immediately sure how.  There's sort of a Catch-22 here because to create the interactivity, we need to specify the variable corresponding to each legend.  To do this, we need to have the legend.  And to have the legend, I think that we need to use `ggplot_build`.  Does that make sense?  Let me know if not.

Also, sorry for the multiple postings, I put all the line notes in the wrong place.

tdhock:
sure, that makes sense that it is necessary to call ggplot_build twice. Can you please add a comment in the code that explains why it is necessary? (next to the second call to ggplot_build)

tdhock:
Also can you please add a test for the new functionality?

I think the iris data example you constructed makes for a very good test example.

Tests for the compiler: test that the list output by animint2dir contains (1) aes(showSelectedfill), (2) selector.types, and (3) first.

Tests for the renderer: start with 150 `<circle>` elements, click on the legend, test for 100 `<circle>` elements, etc. Also test for a size legend, and clicking on that legend does not affect the number of `<circle>` elements.

R/animint.R
```diff
@@ -22,11 +22,54 @@ parsePlot <- function(meta){
...
+          meta$first[[var_name]] <- unique(var)
+        }
+        # if selectortypes is not specified, set it to multiple
+        if(is.null(meta$selector.types[[var_name]])) {
+          meta$selector.types[[var_name]] <- "multiple"
+        }
+      }
+    }
+    ## need to call ggplot_build again because I've added to the plot
+    meta$built <- ggplot2::ggplot_build(meta$plot)
```
kferris10: @tdhock Did you want a comment more explicit than this?

kferris10:
This test should take care of most of what you suggested.  My only question is, what is the easiest way to look at the showSelectedcolour aesthetic?  Unless I'm missing something, `animint2dir` doesn't return it easily.

```
dir <- animint2dir(viz, out.dir = "animint-htmltest", open.browser = F)
str(dir$plots$petal)
List of 15
 $ legend :List of 1
  ..$ Species:List of 6
  .. ..$ guide      : chr "legend"
  .. ..$ geoms      : chr "point"
  .. ..$ title      : chr "Species"
  .. ..$ vars       : chr "Species"
  .. ..$ legend_type: chr [1:2] "colour" "size"
  .. ..$ entries    :List of 3
  .. .. ..$ :List of 4
  .. .. .. ..$ pointcolour: chr "#F8766D"
  .. .. .. ..$ pointsize  : num 1
  .. .. .. ..$ pointshape : num 16
  .. .. .. ..$ label      :Class 'AsIs'  chr "setosa"
  .. .. ..$ :List of 4
  .. .. .. ..$ pointcolour: chr "#00BA38"
  .. .. .. ..$ pointsize  : num 3.5
  .. .. .. ..$ pointshape : num 16
  .. .. .. ..$ label      :Class 'AsIs'  chr "versicolor"
  .. .. ..$ :List of 4
  .. .. .. ..$ pointcolour: chr "#619CFF"
  .. .. .. ..$ pointsize  : num 6
  .. .. .. ..$ pointshape : num 16
  .. .. .. ..$ label      :Class 'AsIs'  chr "virginica"
 $ strips :List of 3
  ..$ top  :List of 1
  .. ..$ : chr ""
  ..$ right:List of 1
  .. ..$ : chr ""
  ..$ n    :List of 2
  .. ..$ top  : num 0
  .. ..$ right: num 0
 $ layout :'data.frame':	1 obs. of  10 variables:
  ..$ PANEL            : int 1
  ..$ ROW              : int 1
  ..$ COL              : int 1
  ..$ SCALE_X          : int 1
  ..$ SCALE_Y          : int 1
  ..$ AXIS_X           : logi TRUE
  ..$ AXIS_Y           : logi TRUE
  ..$ coord_fixed      : logi FALSE
  ..$ width_proportion : num 1
  ..$ height_proportion: num 1
 $ geoms  :List of 1
  ..$ : chr "geom2_point_petal"
 $ axis1  :List of 10
  ..$ x     :List of 3
  .. ..$ : num 0.2
  .. ..$ : num 0.508
  .. ..$ : num 0.816
  ..$ xlab  :List of 3
  .. ..$ : chr "2"
  .. ..$ : chr "4"
  .. ..$ : chr "6"
  ..$ xrange: num [1:2] 0.705 7.195
  ..$ xline : logi FALSE
  ..$ xticks: logi TRUE
  ..$ y     :List of 6
  .. ..$ : num 0.00758
  .. ..$ : num 0.197
  .. ..$ : num 0.386
  .. ..$ : num 0.576
  .. ..$ : num 0.765
  .. ..$ : num 0.955
  ..$ ylab  :List of 6
  .. ..$ : chr "0.0"
  .. ..$ : chr "0.5"
  .. ..$ : chr "1.0"
  .. ..$ : chr "1.5"
  .. ..$ : chr "2.0"
  .. ..$ : chr "2.5"
  ..$ yrange: num [1:2] -0.02 2.62
  ..$ yline : logi FALSE
  ..$ yticks: logi TRUE
 $ xtitle : chr "Petal.Length"
 $ xanchor: chr "middle"
 $ xangle : num 0
 $ ytitle : chr "Petal.Width"
 $ yanchor: chr "end"
 $ yangle : num 0
 $ xlabs  :List of 3
  ..$ : chr "2"
  ..$ : chr "4"
  ..$ : chr "6"
 $ ylabs  :List of 6
  ..$ : chr "0.0"
  ..$ : chr "0.5"
  ..$ : chr "1.0"
  ..$ : chr "1.5"
  ..$ : chr "2.0"
  ..$ : chr "2.5"
 $ title  : chr "Petal Data"
 $ options:List of 2
  ..$ width : num 400
  ..$ height: num 400
```

R/animint.R
tdhock: Please make it as detailed as possible. I would recommend just copy-pasting your comment from above:

```As for calling ggplot_build twice, I'm sure that there is a way around it, but I'm not immediately sure how. There's sort of a Catch-22 here because to create the interactivity, we need to specify the variable corresponding to each legend. To do this, we need to have the legend. And to have the legend, I think that we need to use ggplot_build. Does that make sense? Let me know if not.
```

More generally if you have to make some design decisions it is better to provide justification in the code comments than in the github comments, so it is easier to remember why we made those decisions when we have to come back to the code later.

tdhock:
the aesthetic mapping is geom-specific so is not saved in `plots` (the plot-specific information).

Also, you don't need to call animint2dir if you also are calling animint2HTML (which calls animint2dir and returns all of its information in addition to the rendered HTML).

```
> info <- animint2HTML(viz)
> info$geoms$geom1_point_sepal$aes
                 x                  y             colour               size 
    "Sepal.Length"      "Sepal.Width"          "Species"      "Petal.Width" 
showSelectedcolour 
         "Species" 
```

kferris10:
Got it.  Thanks for your help.

I'll figure out why those tests are failing over the weekend.  After that, is there anything else that we want to add to this feature?

R/animint.R
cpsievert: Do you have a link to that issue you're referring to? Why is having showSelectedfill instead of showSelectedFill an issue?

kferris10: I don't know that it is an issue.  I just wanted to point it out in case there was some preference for making it showSelectedFill, showSelectedColour, etc

tests/testthat/test-interactive-legends.R
```diff
@@ -0,0 +1,58 @@
...
+  geom_point(aes(Sepal.Length, Sepal.Width, 
+                 colour = Species, size = Petal.Width), 
+             data = iris) + 
+  theme(panel.background = element_rect(fill = "lightblue"), 
+        panel.border = element_rect(fill = NA, 
+                                    color = "black", 
+                                    size = 2, 
+                                    linetype = "dashed"), 
+        panel.margin = grid::unit(.1, "cm")) + 
+  facet_wrap(~Species, nrow = 2)
```
cpsievert: I think it's better to not modify the theme here until we merge #84

R/animint.R
```diff
@@ -22,11 +22,59 @@ parsePlot <- function(meta){
...
+      var_name <- plot.meta$legend[[legend.i]]$vars
+      # the actual values for that variable
+      var <- L$data[[var_name]]
+      # checking if it is a discrete variable
+      if(plyr::is.discrete(var)) {
+        # adding a showSelected aesthetic for each legend entry
+        for(i in legend_type) {
+          temp_name <- paste0("showSelected", i)
+          L$mapping[[temp_name]] <- as.symbol(var_name)
+#           browser()
```
cpsievert: I prefer to leave (even commented out) browsers out of the code. It's fine when experimenting, but this should be removed before we merge.

cpsievert:
I agree with @tdhock that it'd be great to have some sort of visual cue that distinguishes selected legend entries from the non-selected ones. 

Although it'd be nice, I don't want it to stall progress, so if it's difficult, let's merge this (once tests are passing) and put it on the TODO list.

kferris10:
These commits fix some of the test failures.  I am working on the rest and hope to be done soon.  

I've [updated the example](https://github.com/tdhock/animint/blob/interactive_legends/inst/examples/interactive-legends.R) to show a potential bug that I came across.  In the example, the sepal plot should only be used for `clickSelects` and these points should not be toggled when they are clicked.  However, because of the way the compiler is currently being modified, I believe that a `showSelected` aesthetic is added to the petal plot (or something to that effect).  [Here's an example](http://bl.ocks.org/kferris10/raw/5671b76aecd04c195d73/).  Click on any of the points in the sepal plot and you will see what I mean.

Should I spend some time trying to fix this?

tdhock:
Interesting example. And good point. Maybe we should tell the compiler to add `aes(showSelected=Species)` only if Species is not specified for any other clickSelects/showSelected aesthetics?

Can you add a test case for the behavior you expect?

kferris10:
Even if I do that, I think that we still might run into issues.  Just for example, how would we expect the world bank example in the facet-lines test work?

kferris10:
Actually, there may be a workaround for this bug with a few changes.

I change it so that instead of a legend adding `showSelectedcolour` aesthetic, it adds `showSelectedlegendcolour`.  Then, when calling `update_selector` by clicking, I only update geoms which have a showSelected aesthetic without legend in the name.  [Working example](http://bl.ocks.org/kferris10/raw/8b56ca7e042f32bdec7a/)

The current code is pretty rough, and I'm not sure if it will work with multiple selection variables or with more than two plots so it may be simpler to just go with Toby's idea.

R/animint.R
```diff
@@ -22,11 +22,58 @@ parsePlot <- function(meta){
...
+      # the name of the variable used in this legend
+      var_name <- plot.meta$legend[[legend.i]]$vars
+      # the actual values for that variable
+      var <- L$data[[var_name]]
+      # checking if it is a discrete variable
+      if(plyr::is.discrete(var)) {
+        # adding a showSelected aesthetic for each legend entry
+        for(i in legend_type) {
+          temp_name <- paste0("showSelectedlegend", i)
+          L$mapping[[temp_name]] <- as.symbol(var_name)
```
kferris10: change name of aesthetic

inst/htmljs/animint.js
```diff
@@ -1458,9 +1458,34 @@ var animint = function (to_select, json_file) {
...
+            // only update the geom if there's a showSelected statement that doesn't come from the legend
+            if(!/legend/.test(entry) && /showSelected/.test(entry)) {
+              update_geom(g_name, v_name);
+            }
+          } else {
+          // otherwise just update the geom as normal
+            update_geom(g_name, v_name);
+          }
+        });
+      });
```
kferris10: when clicking, only update the appropriate geoms

tdhock: s_info.update is an array with the names of the geoms that need to be updated (determined at compile-time).

Why do you need all this new code? it seems rather complicated -- could some of the logic be moved to the compiler?

To ac

tdhock:
about the facet-lines test:

the first geom_line has color=region and clickSelects=country. I would expect that clicking a line changes the selected country and clicking the region legend hides the corresponding lines. https://github.com/tdhock/animint/blob/master/tests/testthat/test-facet-lines.R#L16-L17

for the geom_point in the scatterplot we have color=region, clickSelects=country, and showSelected=year. https://github.com/tdhock/animint/blob/master/tests/testthat/test-facet-lines.R#L22-L25 I would expect again that clicking a point changes the country, and clicking the region legend hides the correpsponding points.

does that clarify your question or am I missing something?

inst/htmljs/animint.js
kferris10: I'm sure that it could be much simpler.  With this example, I was just experimenting with a fix for this bug.

kferris10:
I guess that I don't exactly understand the region-country dynamic.  Let's say that the US is the only country that is selected in the North America region.  When I click on North America in the legend, what should be displayed?

tdhock:
Sorry for confusion about what should be displayed. In fact it is better to think about the selection variables. lets say the viz starts off with

year = 1960
country = ["US", "vietnam"]
region = ["North America", "south america", "europe", ...] (all selected)

click on North america in the legend, and then the selector variables become:

year and country stay the same,
region = ["south america", "europe", ...] (all selected except North America)

so consequently all the lines and points (everything with color=region, which is translated to showSelectedcolor=region) corresponding to countries in North America disappear. However, "US" is still part of the country selection.

kferris10:
That clears it up, thanks!  I think that this should take care of that issue.

Should I write a test for this behavior or is that unnecessary because we'll catch it through the facet-lines test?

tests/testthat/test-legend-interactivity.R
```diff
@@ -0,0 +1,58 @@
...
+})
+
+test_that("clicking sepal.width legend does nothing", {
+  clickID("2.5")
+  expect_equal(length(get_circles("sepal")), 150)
+  expect_equal(length(get_circles("petal")), 150)
+})
+
+test_that("clicking Sepal point doesn't affect sepal plot", {
+#   clickID("thecircleiwanttoclick")
```
kferris10: To run this test, I need to click on one of the circles in the plot.  Is `clickID` the appropriate way to do that?

tdhock: yes!

tdhock:
Thanks for proposing to add a test for the interactive legend features in the WorldBank data viz. For example give `aes(id=country)` to the lines and then test that United States disappears when you click on the North america legend. Yes I think that would be a good idea. Either add some code to the end of one of the existing test files (test-widerect.R for example) or add a new test, whichever you prefer.

tests/testthat/test-legend-interactivity.R
kferris10: Is there an example of how to do this somewhere?  None of the circles have IDs so I'm a bit confused.

kferris10:
Is there an example somewhere of you guys doing something similar?  I'm getting a little bit stuck trying to implement this.

tdhock:
is the problem that the `id` property is not show up in the `<path>` elements that correspond to the `geom_line`? 

what exactly is the problem you are having?

tdhock:
really great work Kevin, this is exactly the kind of interactivity that I imagined for the WorldBank viz. You made it a reality!

R/animint.R
tdhock: the test that is currently failing (test-panels.R, "renderer can handle no grid lines" `Error: length(first) == 1 is not TRUE`) is because the sex_smoker selector was inferred to be single selection before this. 

Careful: `meta$selector.types$sex_smoker` is the user-specified value (undefined in at the beginning of this test but set to "multiple" by your code), and `meta$selectors$sex_smoker$type` is the inferred type which is actually saved to plot.json and used by the renderer.

kferris10:
So is there anything else that we want to do here or is this ready to go?

tdhock:
I will try to implement the fading legend titles tomorrow, I don't think it
should be too hard.

kferris10:
Hey Toby, I accidentally pushed to this branch.  I believe I reset it back to your latest commit, but could you confirm that I did this correctly?

tdhock:
yeah @kferris10 it looks like your reset was fine.

I implemented opacity of 0.5 for legend entries.

however I discovered a bug that hides the "year = 2003" geom_text when you click on the "status = odd" legend entry. I already added a test but I don't have time to fix it right now. I think the compiler is just accidentally adding a showSelectedlegend aes to that geom_text.

R/animint.R
```diff
@@ -22,11 +22,76 @@ parsePlot <- function(meta){
...
+        if(!var.is.interactive){
+          for(i in legend_type) {
+            temp_name <- paste0("showSelectedlegend", i)
+            L$mapping[[temp_name]] <- as.symbol(var_name)
+          }
+        }
+        # if selector.types is not specified, set it to multiple
+        if(is.null(meta$selector.types[[var_name]])) {
+          meta$selector.types[[var_name]] <- "multiple"
+        }
```
kferris10: @tdhock Would it make sense to coerce `meta$selector.types` to a list?  I'm getting an error when trying to run this example.

```
mtcars$cyl <- factor(mtcars$cyl)
mtcars$vs <- factor(mtcars$vs)
p1 <- ggplot() + 
  geom_point(aes(wt, mpg, colour = cyl, size = vs), 
             data = mtcars)
p2 <- ggplot() + 
  geom_point(aes(wt, drat, colour = cyl, size = vs, clickSelects = vs), 
             data = mtcars)
animint2dir(list(p1 = p2, p2 = p2))
```

kferris10:
You're right about the bug being caused by accidentally adding a showSelected aesthetic to the text.  It's also happening to the widerects and tallrects as well.  The easy solution would be to simply tell the compiler not to add the showSelected aesthetic for `geom_widerect()`, `geom_tallrect()`, or `make_text()`.  

Would this be an okay solution?  I can't think of a time when we'd want to use the legend to hide these.

tdhock:
Actually I think the correct behavior is indeed to show/hide the
wide/tallrects after clicking the status legend, since they have
aes(linetype=status). However the geom_text does not have
aes(linetype=status) so it should not show/hide after clicking the status
legend.

And yes it would be fine to make selector.types a list if you like.

kferris10:
I think that bug is now taken care of.  What else do we want to work on here?

cpsievert:
This is so awesome! I'll take a closer look this afternoon.

I'd be nice to have a visual cue that the legends are click-able. Maybe set the [cursor property of the legend entires to be "pointer"?](http://www.w3schools.com/jsref/prop_style_cursor.asp)

cpsievert:
@kferris10 I see:

```r
> animint2dir(list(p = qplot(Sepal.Length, Sepal.Width, colour = Species, data = iris)))
Error in L$data[[var_name]] : attempt to select less than one element
```

Is the intention to have clickable legends for this single plot example ([as Toby mentioned](https://github.com/tdhock/animint/pull/74#issuecomment-104728425))? If so, please add an appropriate test.

kferris10:
@cpsievert I think that this is due to the way which animint handles `qplot()`.  This works fine for me `animint2dir(list(p = ggplot() + geom_point(aes(Sepal.Length, Sepal.Width, colour = Species), data = iris)))`

I'll check this out

kferris10:
So I believe that I need to update the compiler so that it compiles differently for plot created using `qplot()` and those using `ggplot() + geom_point()` because I need to ask the data in the ggplot to create the `showSelectedlegend` aesthetic.

If I run the example from before, and insert a `browser()` call at line 38, the data are located in `meta$plot$layers` and not in `meta$plot$data`
```
animint2dir(list(p = ggplot() + geom_point(aes(Sepal.Length, Sepal.Width, colour = Species), data = iris)))

Browse[1]> str(meta$plot$layers[[1]]$data)
'data.frame':	150 obs. of  5 variables:
 $ Sepal.Length: num  5.1 4.9 4.7 4.6 5 5.4 4.6 5 4.4 4.9 ...
 $ Sepal.Width : num  3.5 3 3.2 3.1 3.6 3.9 3.4 3.4 2.9 3.1 ...
 $ Petal.Length: num  1.4 1.4 1.3 1.5 1.4 1.7 1.4 1.5 1.4 1.5 ...
 $ Petal.Width : num  0.2 0.2 0.2 0.2 0.2 0.4 0.3 0.2 0.2 0.1 ...
 $ Species     : Factor w/ 3 levels "setosa","versicolor",..: 1 1 1 1 1 1 1 1 1 1 ...

Browse[1]> str(meta$plot$data)
 list()
 - attr(*, "class")= chr "waiver"
```

If, however, I change to using `qplot()`
```
animint2dir(list(p = qpot(Sepal.Length, Sepal.Width, colour = Species, data = iris)))

Browse[1]> str(meta$plot$layers[[1]]$data)
 list()
 - attr(*, "class")= chr "waiver"

Browse[1]> str(meta$plot$data)
'data.frame':	150 obs. of  5 variables:
 $ Sepal.Length: num  5.1 4.9 4.7 4.6 5 5.4 4.6 5 4.4 4.9 ...
 $ Sepal.Width : num  3.5 3 3.2 3.1 3.6 3.9 3.4 3.4 2.9 3.1 ...
 $ Petal.Length: num  1.4 1.4 1.3 1.5 1.4 1.7 1.4 1.5 1.4 1.5 ...
 $ Petal.Width : num  0.2 0.2 0.2 0.2 0.2 0.4 0.3 0.2 0.2 0.1 ...
 $ Species     : Factor w/ 3 levels "setosa","versicolor",..: 1 1 1 1 1 1 1 1 1 1 ...
```

Just wanted to check and see if you guys had any thoughts before I got started on this.

cpsievert:
Maybe you could just fall back on the `meta$plot$data` if `meta$plot$layers[[i]]$data` doesn't exist (for each layer i)? So something like:

```r
dat <- with(meta$plot, if (length(layers[[i]]$data) > 0) layers[[i]]$data else data)
```

R/animint.R
```diff
@@ -4,6 +4,14 @@
...
 #' @import ggplot2
 parsePlot <- function(meta){
+  ## adding data and mapping to each layer from base plot, if necessary
+  for(layer.i in seq_along(meta$plot$layers)) {
+    ## if data is not specified, get it from plot
+    if(length(meta$plot$layers[[layer.i]]$data) == 0) meta$plot$layers[[layer.i]]$data <- meta$plot$data
+    ## if mapping is not specified, get it from plot
+    if(is.null(meta$plot$layers[[layer.i]]$mapping)) meta$plot$layers[[layer.i]]$mapping <- meta$plot$mapping
+  }
+
```
kferris10: Actually, this is a much easier solution.  I just add the data and mapping from the overall plot to each layer if necessary.  Now, both `qplot(Sepal.Length, Sepal.Width, colour = Species, data = iris)` and `ggplot(iris, aes(Sepal.Length, Sepal.Width, colour = Species)) + geom_point()` work.  Does anybody foresee any potential complications with this approach?

tdhock:
so is this PR ready to merge?

cpsievert:
@tdhock it seems mostly ready to me, but I'd prefer to merge # 95 first (if we don't, the build will fail because [of this](https://github.com/tdhock/animint/pull/97#issuecomment-120951216))

kferris10:
So is everybody happy with this PR?

tdhock:
I got an error when running the tests from the command line

```
Error in loadNamespace(name) (from test-panels.R#22) : there is no package called ‘ggthemes’
```

not a deal breaker but I thought I would mention it.

tdhock:
I also got a test failure on firefox

```
1. Failure (@test-panels.R#87): grid lines are drawn correctly -----------------
toupper(value) not equal to toupper(expected_string)
1 string mismatches:
x[1]: "RGB(255, 255, 255)"
y[1]: "RGB(229, 229, 229)"
```

tdhock:
and in one of the shiny tests

```
/home/thocking/.cabal/bin/pandoc /tmp/RtmpxPseip/index.utf8.md --to html --from markdown+autolink_bare_uris+ascii_identifiers+tex_math_single_backslash-implicit_figures --output /tmp/RtmpxPseip/file2d54715606e4.html --smart --email-obfuscation none --standalone --section-divs --template /home/thocking/lib/R/library/rmarkdown/rmd/h/default.html --variable 'theme:bootstrap' --mathjax --variable 'mathjax-url:https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML' --no-highlight --variable highlightjs=file2d54715606e4_files/highlight 

Output created: /tmp/RtmpxPseip/file2d54715606e4.html
Error in time.classes[[1]] : subscript out of bounds
In addition: Warning message:
In normalizePath(file, winslash = "/") :
  path[1]="index.Rmd": No such file or directory
2. Failure (@test-shiny.R#37): animint plot renders in an interactive document -
length(circles) >= 1 isn't true
```

tdhock:
however i am not sure if these test failures are also happening on master, let me check...

tdhock:
on master I get this error but not the other one

```
1. Failure (@test-panels.R#87): grid lines are drawn correctly -----------------
toupper(value) not equal to toupper(expected_string)
1 string mismatches:
x[1]: "RGB(255, 255, 255)"
y[1]: "RGB(229, 229, 229)"
```

tdhock:
The color mismatch error is because I set `theme_bw()` instead of the default `theme_grey()` in my ~/.Rprofile -- can we add `theme_set(theme_grey())` to the testthat.R file if we depend on those theme colors in the tests?

Are you guys seeing the shiny test failure on your machines?

cpsievert:
@tdhock tests pass on my machine...does the shiny app show up for you in firefox? I've had problems with shiny tests randomly failing in the past and for that reason [we actually skip it on Travis](https://github.com/tdhock/animint/blob/21dddc8503e9daa2e492ef313d84c5e02fd2430a/tests/testthat/test-shiny.R#L3-L5)

kferris10:
Sometimes the shiny tests fail for me, and sometimes they don't.  They seem to be working fine now.

@tdhock I'll fix the other two problems you noted.  It shouldn't take much time to get them out of the way.

tests/testthat/test-panels.R
```diff
@@ -3,6 +3,7 @@ context("Panel background")
 p1 <- ggplot() +
   geom_point(aes(Sepal.Length, Sepal.Width,
                  colour = Species, size = Species), data = iris) +
+  theme_grey() +
```
kferris10: @tdhock This should take care of the `theme_bw()` issue

tdhock: ok well if the shiny test works on your machines then that is fine with me
+1

kferris10:
@cpsievert [TADA](http://bl.ocks.org/kferris10/raw/65344ee0394bb66c4174/)

cpsievert:
excited_beyond_belief

kferris10:
This example is creating some major problems, but it's also a pretty stupid example.
`p <- qplot(1:10, 1:10, colour = rep(c("red", "blue"), 5))`

cpsievert:
Great work on this @kferris10! Go ahead and bump the version, update NEWS, and merge!

(that example doesn't seem to work on master either, so I'm not too worried about it)

PR #76, opened 2015-05-22
optimizations to save disk space

tdhock:
I ran into some problems while implementing an animint with 206 selection variables http://bl.ocks.org/tdhock/raw/36dc08ace6bd39f8693e/ source: https://github.com/tdhock/animint-examples/blob/master/examples/PSJ.R

I used a for loop to define a bunch of layers with small data sets, each with a different `aes_string` mapping. This results in a lot of wasted disk space: 1180 geoms split into 1767 files for a total of 8.3M on my ext4 filesystem.

Just looking at the tsv files there are many which are under 1KB each, for a rather reasonable total of about 1MB:
```
~/R/animint-examples/examples $ du --block-size=1 -cs --apparent-size PSJ-for/*.tsv|tail
480	PSJ-for/geom98_segment_coverage_chunk1.tsv
154	PSJ-for/geom99_segment_coverage_chunk1.tsv
261	PSJ-for/geom99_segment_coverage_chunk2.tsv
368	PSJ-for/geom99_segment_coverage_chunk3.tsv
475	PSJ-for/geom99_segment_coverage_chunk4.tsv
153	PSJ-for/geom9_segment_coverage_chunk1.tsv
259	PSJ-for/geom9_segment_coverage_chunk2.tsv
365	PSJ-for/geom9_segment_coverage_chunk3.tsv
471	PSJ-for/geom9_segment_coverage_chunk4.tsv
961044	total
```

but actually if you count the filesystem meta-data the data files take up a lot more space:
```
~/R/animint-examples/examples $ du --block-size=1 -cs PSJ-for/*.tsv|tail
4096	PSJ-for/geom98_segment_coverage_chunk1.tsv
4096	PSJ-for/geom99_segment_coverage_chunk1.tsv
4096	PSJ-for/geom99_segment_coverage_chunk2.tsv
4096	PSJ-for/geom99_segment_coverage_chunk3.tsv
4096	PSJ-for/geom99_segment_coverage_chunk4.tsv
4096	PSJ-for/geom9_segment_coverage_chunk1.tsv
4096	PSJ-for/geom9_segment_coverage_chunk2.tsv
4096	PSJ-for/geom9_segment_coverage_chunk3.tsv
4096	PSJ-for/geom9_segment_coverage_chunk4.tsv
7471104	total
```
So the data files take up almost 8MB of disk space, meaning 7MB just of meta-data.

I propose that by default the compiler should not split any geoms into multiple chunks of less than 4KB. In the example above, geom9 was split into 4 chunks but it should not be split (just 1 chunk).

I have already implemented a test using the breakpoints data.


tdhock:
another optimization that I may not have time to implement right away is to reduce the number of geoms from 1180 (using for loops) to only 15 (using .variable and .value aesthetics), which will probably result in a significant decrease from 1767 tsv files.

tdhock:
On the travis build machine the meta-data takes up 4KB as well. https://travis-ci.org/tdhock/animint/builds/63717045#L1469

tests/testthat/test-chunks.R
```diff
@@ -40,3 +40,74 @@ test_that("produce informative errors for bad chunk_vars", {
...
+for(f in test.paths){
+  unlink(f)
+  cat("foo", file=f)
+}
+du.bytes <- bytes.used(test.paths)
+apparent.bytes <- bytes.used(test.paths, "--apparent-size")
+byte.df <- data.frame(du.bytes, apparent.bytes,
+                      file.size=file.size(test.paths),
+                      test.paths)
+print(byte.df)
```
cpsievert: @tdhock could you add a check for `du` (possibly via `Sys.which("du")`) and exit gracefully if it isn't found?

cpsievert:
Looking good @tdhock!

> I propose that by default the compiler should not split any geoms into multiple chunks of less than 4KB. 

Is there anything special about that 4KB number? Maybe the filesystem attaches that much meta-data to *any* tsv file? Any idea if that is a common thing for other file systems to do?

tests/testthat/test-chunks.R
tdhock: OK, I added a `tryCatch` since `system()` stops with an error when du is not found.

tdhock:
I don't know if it is common to have 4KB of meta-data, but I guess it should be filesystem-dependent.

Yes the filesystem attaches 4KB of meta-data to any tsv file (and in fact any file).

But even if 4KB is a heuristic, it certainly is useful. I just re-compiled the data viz with 206 selectors and observed the following results:

| version      | geoms | tsv files |      du | du --apparent-size |
|--------------+-------+-----------+---------+--------------------|
| [[http://bl.ocks.org/tdhock/raw/c505b96ec93e1006044f/][no 4KB limit]] |   670 |      1760 | 7471104 |             961044 |
| [[http://bl.ocks.org/tdhock/raw/cfe03989e8d68f444206/][4KB limit]]    |   670 |       670 | 3166208 |            1011090 |

tdhock:
There is a bug in the renderer for http://bl.ocks.org/tdhock/raw/cfe03989e8d68f444206/ -- we should not download all the data at first. In particular the modelSelection plot should show and download nothing until the user clicks on one of the problems.

tdhock:
@caijun please look at this re-design of the FluView data viz http://bl.ocks.org/tdhock/raw/1df80a2f58da037de923/ source: https://github.com/tdhock/AnimintTest/blob/master/test.easy.R Basically I just replaced the `geom_polygon` with a `geom_point`.

The points do not change position with different values of showSelected, so it would be nice if the compiler changes that we discussed today could also apply to this case. In particular, even though this geom_point does not have the "group" aesthetic, we could create one inside the compiler, so that we divide the data into two pieces:

1. a data.frame with columns x, y, group, showSelected, clickSelects (the information which does not change between showSelected subsets).

2. a data.frame with columns group, fill (the information which does change between showSelected subsets).

caijun:
I looked at your `geom_point` version of the FluView viz. Though it is still a bit slow, it is much faster than the `geom_polygon` version as the file size of each stateMap chunk is much reduced. In your version of `geom_point`, As one weekend corresponds to one state map and one point corresponds to one state, I expect that there are only `48` rows of fill field for each stateMap chunk, while there are actually `48*48=2304` rows in stateMap chunk. I don't know the reason for such a structure of stateMap chunk.

See the simpler example http://bl.ocks.org/caijun/raw/e3c72a9692bd744c5058/, from which I removed the third plot for easier testing.

I added [test-save-separate-chunks.R](https://github.com/tdhock/animint/blob/big-chunks/tests/testthat/test-save-separate-chunks.R) to `big-chunks` branch. In the future, we will work on this test example.

tdhock:
I think there is a bug in your new viz with only 2 plots http://bl.ocks.org/caijun/raw/e3c72a9692bd744c5058/

The `<circle>` elements are always the same color, and there are too many of them (there should be only 48 at any given time).

tdhock:
For the test I guess you should check how many chunk tsv files are produced by animint2dir.

Also I think it would be good to test both the `geom_point` version and the `geom_polygon` version.

tdhock:
also about the slowness I made a static (pure HTML+SVG, no JavaScript) version of the FluView viz http://bl.ocks.org/tdhock/raw/86d25657f8d54246841d/ source: https://gist.github.com/tdhock/86d25657f8d54246841d

the fact that it still has slow scrolling (in firefox on my computer) tells me that there is nothing we can do to our animint.js renderer code to speed up the rendering. Plotting >10,000 SVG elements causes an inherent slowdown to the web browser.

caijun:
You're right. I made a mistake in the two plots version of FluView. I have fixed it and updated the test example. Now the rows for each chunk tsv and the number of chunks are within my expectation for both `geom_point` and `geom_polygon` versions: `347` chunk TSVs as there are `347` weeks, and `48` rows for `geom_point` version as there are `48` states, `15527` rows for `geom_polygon` version as there are `15527` coordinates for plotting the borders of `48` states.

caijun:
As for your static version of the FluView viz, it doesn't respond for scrolling clicks on either my chrome or my safari browser. The viz almost get my firefox  stuck.

The uploaded test of either `geom_point` or `geom_polygon` version of FluView viz has no slowness of scrolling clicks on the top plot. Once you click for showing selected map, it only takes some time to load the chunk TSV for displaying, which gives a feeling of slowness for the viz . This is why I think the file size of chunk TSV is the key to optimization.

tdhock:
it is normal that the static version does not respond to mouse clicks, since I deleted all of the JS code. That implies that at least on firefox, our animint.js code is not responsible for the slowness.

tests/testthat/test-save-separate-chunks.R
```diff
@@ -0,0 +1,72 @@
+context("save chunks")
+
+library(cdcfluview)
+# retrieve state-level data from the CDC's FluView Portal
+state_flu <- get_state_data(2008:2014)
+# data clean
```
tdhock: Rather than downloading the data during each run of this test, please just download it once and store it as a list called `FluView` that contains all the data that you need for this data viz. Then do `save(FluView, file="animint/data/FluView.RData")` and in this test do

```r
data(FluView)
```

R/animint.R
```diff
@@ -557,34 +557,78 @@ saveLayer <- function(l, d, meta){
...
+        ## be downloaded at first if it is not needed.
+        not.one <- subset(selector.df, 1 < chunks.for)
+        if(nrow(not.one) == 0){
+          NULL
         }else{
-          nest.cols <- subset.vec
-          chunk.cols <- NULL
+          rownames(not.one)[[which.max(not.one$min.bytes)]]
         }
       }
```
tdhock: by default the compiler now checks for showSelected variables that would create only 1 tsv file after splitting, and definitely includes them in chunk_order.

This fixes the bug with downloading too many chunks when the viz is first rendered.

tests/testthat/test-variable-value.R
```diff
@@ -0,0 +1,178 @@
...
+##animint2dir(viz.for, "problem-peaks-for")
+
+test_that("some chunks are not downloaded", {
+  info <- animint2HTML(viz.for)
+  node.set <-
+    getNodeSet(info$html, '//td[@class="downloaded"]')
+  value.vec <- sapply(node.set, xmlValue)
+  expect_true("0" %in% value.vec)
+  expect_match(value.vec,  "[01]")
+})
```
tdhock: this test makes sure that some chunks are not downloaded at first (0) and that those which are downloaded only have 1 chunk.

tdhock:
the big data viz now loads a lot faster, since it only downloads the tsv chunks necessary for the first display http://bl.ocks.org/tdhock/raw/b77c1a7e4d6aee40bf6c/

R/animint.R
```diff
@@ -649,12 +694,26 @@ saveLayer <- function(l, d, meta){
...
+          chunk.shape <- file.path(meta$out.dir, csv.name)
+          if(!file.exists(chunk.shape)){
+            x1 <- x[c("x", "y", "group")]
+            write.table(x1, chunk.shape, quote = FALSE, row.names = FALSE, 
+                        sep = "\t")
+          }
+          # info for each group
+          x2 <- x[, !names(x) %in% c("x", "y")]
+          x <- x2[!duplicated(x2), ]
+          }
```
tdhock: Can you please use more informative variable names than `x1`, `x2`?

For example, `chunk.specific` instead of `x2` and `same.in.all.chunks` for `x1`.

Also, do you think it is necessary to check that the geom is a line/path/polygon/ribbon? I think this is OK for a first attempt, but eventually we should use a more relevant test: does this geom have any repeated data? (maybe not necessarily just x,y columns)

caijun: I think the redundancy mostly occurs when drawing spatial objects (excluding point). For non-spatial objects, it may occur the join relation of n:1 for constant fields and updating fields. If I split variable `x` based on whether repeated data present, how do I deal with the case that the updating field indeed has duplicated values, for example in my `geom_polygon` example, some states are of the same fill colors.

Besides, I am not clear that what should I test if write a test for this function? Could you give a hint?

tdhock: for example in the WorldBank data viz we have a geom_point with repetitive data columns that are the same for each country in each chunk tsv file: color, clickSelects. However the x and y columns are different in each tsv file.

tdhock: About the case that the updating field indeed has duplicated values, there should be some way to detect that x,y columns are the same across each showSelected subset, and that fill is different.

tdhock: I think a good test would be for presence of x y columns in geomXX_shape.tsv, and absence of x y columns in each of the geomXX_chunkXYZ.tsv files.

tdhock: finally I think it would significantly simpler to save the shape.tsv file outside of the saveChunks function -- do you think it would be possible to move this code inside saveLayer, just before calling saveChunks? Something like

```r
differences.between.chunks <- saveDataSameInAllChunks(g.data, chunk.cols)
g$chunks <- saveChunks(differences.between.chunks, chunk.cols, meta)
```

caijun: Firstly I would like to save the extra shape.tsv outside of `saveChunks` and directly utilize `saveChunks`, but I found the filename of chunks could be only be modified in the inside of `saveChunks`. To modify the code as less as possible, I just put them inside `sideChunks`. To put shape.tsv outside of `saveChunks` can avoid to creat `meta$geom`. Moving the code inside saveLayer is possible.

caijun: yeah, the absence of x y columns in each of the geomXX_chunkXYZ.tsv files is what I should test and the number of rows should be less than that of `g.data`.

caijun: I look at the WorldBank data viz. The `color` and `clickSelects` are the same for  in each chunk tsv file. I wonder whether it's worth to reduce output chunk file size since this small redundancy make the renderer render the viz much easier. I think I should make the `geom_polyon` example work for both compiler and render, then I will extend it to WorldBank data viz.

tdhock:
I have implemented compiler and renderer support for the .variable and .value aesthetics. For this example http://bl.ocks.org/tdhock/raw/cfc5d67636ee76d42be0/ (source:
https://github.com/tdhock/animint-examples/blob/master/examples/PSJ.R)
there are big improvements in construction and compile times

```
constructing data viz with for loops
    user  system elapsed 
 38.356   0.104  38.769 

compiling data viz
   user  system elapsed 
 26.944   0.348  27.520 

constructing data viz with .variable .value
   user  system elapsed 
  0.112   0.000   0.114 
 
compiling data viz
   user  system elapsed 
  1.204   0.004   1.220 
```

as well as disk space

```
thocking@silene:~/R/animint-examples/examples(master)$ du -hs PSJ-for-new/ PSJ
4.0M	PSJ-for-new/
1.7M	PSJ
```

This is mainly because the for loop version has many small files (one for each geom), with 670 tsv files total:

```
thocking@silene:~/R/animint-examples/examples(master)$ du -hsc PSJ-for-new/*.tsv|tail
4.0K	PSJ-for-new/geom92_segment_coverage_chunk1.tsv
4.0K	PSJ-for-new/geom93_segment_coverage_chunk1.tsv
4.0K	PSJ-for-new/geom94_segment_coverage_chunk1.tsv
4.0K	PSJ-for-new/geom95_segment_coverage_chunk1.tsv
4.0K	PSJ-for-new/geom96_segment_coverage_chunk1.tsv
4.0K	PSJ-for-new/geom97_segment_coverage_chunk1.tsv
4.0K	PSJ-for-new/geom98_segment_coverage_chunk1.tsv
4.0K	PSJ-for-new/geom99_segment_coverage_chunk1.tsv
4.0K	PSJ-for-new/geom9_segment_coverage_chunk1.tsv
3.1M	total
thocking@silene:~/R/animint-examples/examples(master)$ ls PSJ-for-new/*.tsv|wc -l
670
```

in contrast the .variable .value version uses only 27 tsv files:

```
thocking@silene:~/R/animint-examples/examples(master)$ ls PSJ/*.tsv|wc -l
27
thocking@silene:~/R/animint-examples/examples(master)$ du -hsc PSJ/*.tsv|tail
16K	PSJ/geom6_tallrect_coverage_chunk4.tsv
16K	PSJ/geom6_tallrect_coverage_chunk5.tsv
16K	PSJ/geom6_tallrect_coverage_chunk6.tsv
16K	PSJ/geom6_tallrect_coverage_chunk7.tsv
16K	PSJ/geom6_tallrect_coverage_chunk8.tsv
16K	PSJ/geom6_tallrect_coverage_chunk9.tsv
276K	PSJ/geom7_segment_coverage_chunk1.tsv
108K	PSJ/geom8_segment_coverage_chunk1.tsv
4.0K	PSJ/geom9_tallrect_resError_chunk1.tsv
1.2M	total
```

tdhock:
@caijun can you please take a look at why some tests are failing, and try to fix them? https://travis-ci.org/tdhock/animint/builds/66231652#L1585

caijun:
ok, I am working on it.

tdhock:
I guess it has something to do with the fact that you have changed the animint.R compiler and not the animint.js renderer.

R/animint.R
caijun: Actually in the WorldBank data viz, the `color` and `clickSelects` columns are different for chunks. For example, the `colour` are different at `78  79  81  84  85  86  87  88  89  90  91  93  94  95  96  97  98  99 101` rows for the first two chunks. @tdhock could you come up with another example for the case you mentioned?

tests/testthat/test-chunks.R
```diff
@@ -40,3 +40,78 @@ test_that("produce informative errors for bad chunk_vars", {
...
+         geom_line(aes(segments, error, group=bases.per.probe,
+                       clickSelects=bases.per.probe),
+                   data=only.error, lwd=4))
+
+bytes.used <- function(file.vec, ...){
+  ## Note: the --apparent-size flag gives sizes that are consistent
+  ## with file.info, but those sizes actually under-estimate the
+  ## actual amount of space used on disk.
+  file.str <- paste(file.vec, collapse=" ")
+  cmd <- paste("du --block-size=1", ..., file.str)
```
caijun: This usage of `du` produces following error when I run `test_run()` on My Mac OS X Yosemite 

```
du: illegal option -- -

usage: du [-H | -L | -P] [-a | -s | -d depth] [-c] [-h | -k | -m | -g] [-x] [-I mask] [file ...]

Error in data.frame(du.bytes, apparent.bytes, file.size = file.size(test.paths),  : 
  arguments imply differing number of rows: 0, 3
```

tdhock: go ahead and change the du code so that it works on your computer, or just comment it out.

R/animint.R
tdhock: Which version of the WorldBank viz did you look at?

For the first example wb.all, there is a row for every (year,country)
combination
https://github.com/tdhock/animint/blob/master/inst/examples/WorldBank.R#L4
so each chunk has the same number of rows, and I guess it should be easy to
detect that color/clickSelects are the same across chunks.

For the other examples which use the not.na data set there are some rows
that have been removed, so each chunk does not have the same number of
rows, and I guess it would not be feasible to detect.

caijun: Oh, there isn't `wb.all` example in my WorldBank viz. Thanks for your information, I will use the one you provided.

tdhock:
@caijun you said "It reduces the output file size from 50.5 MB to 49.3 MB" but what example is that referring to? Can you please post the source code you used to create both versions? It seems to me strange that the WorldBank viz should take up so much disk space. On my computer I see 10 times less disk usage.

```
thocking@silene:~$ du -hs ~/R/animint/inst/examples/WorldBank-all/
4.5M	/home/thocking/R/animint/inst/examples/WorldBank-all/
```

 I did `devtools::install_github("tdhock/animint@7e57b1f334327fc4497ced1d833888c047875a70")` (current master) and then ran the first viz in https://github.com/tdhock/animint/blob/master/inst/examples/WorldBank.R#L1-L32.

caijun:
I create `WorldBank-all-new` and `WorldBank-all-old` using following R code

https://github.com/caijun/AnimintTest/blob/master/dev/test-WorldBank-all.R

The output filesize is as follows
```
TTMBP:animint tonytsai$ du -hs WorldBank-all-new/
 47M	WorldBank-all-new/
TTMBP:animint tonytsai$ du -hs WorldBank-all-old/
 48M	WorldBank-all-old/
```

I use this WorldBank-all viz to test multiple vars selection for `saveCommonChunk`. The difference between mine and yours is that I force multiple vars in the WorldBank data viz by specifying

```
geom_text(aes(life.expectancy, fertility.rate, label=country,
                     showSelected=country, showSelected2=year,
                     key=country), #also use key here!
                 data=WorldBank, chunk_vars=c("year", "country"))
```

See above comments.

tdhock:
ok that makes sense. thanks for clarifying what exactly you computed and why it was so big.

tdhock:
do you need any help with modifying the renderer code?

tests/testthat/test-save-separate-chunks.R
```diff
@@ -0,0 +1,231 @@
...
+                clickSelects=country),
+            data=WorldBank, size=4, alpha=3/5)
+
+bar=ggplot()+
+  theme_animint(height=2400)+
+  geom_bar(aes(country, life.expectancy, fill=region,
+               showSelected=year, clickSelects=country),
+           data=WorldBank, stat="identity", position="identity")+
+  coord_flip()
+
```
tdhock: for a simpler test you could delete the bar plot from each of these viz.

tests/testthat/test-save-separate-chunks.R
```diff
@@ -0,0 +1,231 @@
...
+  merge(USdots, df, by.x = "region", by.y = "state")
+})
+
+test_that("save separate chunks for geom_point without group", {
+  # the compiler will not break a geom into chunks if any of the resulting 
+  # chunk tsv files is estimated to be less than 4KB.
+  state.map <- p + 
+    geom_point(data = map_flu, aes(x = mean.long, y = mean.lat, fill = level, 
+                                   showSelected = WEEKEND), color = "black", size = 10)
+  viz <- list(levelHeatmap = level.heatmap, stateMap = state.map, title = "FluView")
```
tdhock: it would make for a simpler test if yo combined this test "geom_point without group" with the above test "geom_polygon" -- all you would have to do is add the geom_point to the above plot. And then you could test both the polygon tsv files and the point tsv files.

tests/testthat/test-save-separate-chunks.R
```diff
@@ -0,0 +1,231 @@
...
+  varied.chunks <- list.files(path = out.dir, pattern = "geom.+point.+chunk[0-9]+.tsv", 
+                              full.names = TRUE)
+  # number of chunks
+  expect_equal(length(common.chunk), 1L)
+  no.chunks <- length(varied.chunks)
+  expect_equal(no.chunks, length(unique(map_flu$WEEKEND)))
+  # test common.chunk
+  common.data <- read.csv(common.chunk, sep = "\t")
+  expect_equal(nrow(common.data), nrow(USdots))
+  expect_true(all(c("x", "y") %in% names(common.data)))
```
tdhock: you don't expect "group" to be a column name in the geom_point tsv files?

in that case to combine the comomn tsv file with the chunk tsv files, you expect that there are exactly the same number of rows and they are in the same order?

tests/testthat/test-save-separate-chunks.R
caijun: for the test "geom_point without group", I test two situations: one is not to break `geom_point` into chunks due to the resulting chunks less than 4KB, the other is to force to split `geom_point` into chunks without regard for chunk file size. The testing objective is much different from that of `geom_polygon` and I prefer not ot combine them into one test.

tests/testthat/test-save-separate-chunks.R
caijun: yes, since the input data `map_flu` doesn't have a `group` column. But if it presents, it also works as the `group` column has been tested in  `geom_polygon` test. Without `group` column, I split the data.frame with the assumption that there are no other operations like  `sort` to change the order of observations. The number of rows in common tsv and varied tsvs are the same.

caijun:
I have no time to work on the modification of renderer  since I have been preparing my Ph.D. qualifying exam and thesis proposal in June 26. You can provide me some function prototypes to implement in renderer. After June 26, I can implement them as soon as possible. For these days, I can only respond to    the compiler, which part I am much more familiar with.

tdhock:
Hey @caijun did you revert the commits that I pushed that implemented the .variable .value aesthetics? Would you prefer I push them to a separate branch?

caijun:
Hi Toby. I passed PhD qualifying exam. I am back to animint. Once I reverted your commits when I tried to reverted my commits. yeah, I can be more focused if you can push your commits to a separate branch.

tdhock:
congrats on passing your qualifying exam!

I won't be able to push any commits in the next 2 weeks (away from the
office on a work trip to France), but I will think about creating a new
branch after I get back.

tdhock:
this is a good first implementation Tony! Glad to see the FluView viz is small enough to upload as a gist now.

About the speed I would suggest not working on it in this PR. Instead can you please take a look at the tests that are failing and see if you can fix them? Then we can merge this PR with master, and later you can submit another PR about speeding up the rendering. Is that OK with you?

caijun:
I agree with you. This PR contains too much modifications. We should end end it at this point. I will try to fix those failed tests and merge this PR.

caijun:
I am trying to fix the failed `test-facet-trivial.R`. I add a breakpoint to following line

https://github.com/tdhock/animint/blob/big-chunks/R/animint.R#L39

```
plot.meta$strips <- with(meta$built, getStrips(plot$facet, panel))
```

In `big-chunk` branch, `plot.meta$strips` is

```
Browse[2]> plot.meta$strips
$right
[1] "plantae"

$top
[1] "plantae"

$n
$n$top
[1] 1

$n$right
[1] 1
```

While in `master` branch, `plot.meta$strips` is

```
Browse[2]> plot.meta$strips
$right
$right[[1]]
[1] "plantae"


$top
$top[[1]]
[1] "plantae"


$n
$n$top
[1] 1

$n$right
[1] 1
```

Since I didn't modify anything about `meta$built` and `getStrips` function, it's weird for me why `plot.meta$strips` has been changed. @tdhock could you give me a hint that is there any other part of `animint.R` that could modify strips? I am still confusing how this change occurres.

tdhock:
I'm not sure why plot.meta$strips would change either.

Anyways test-facet-trivial.R does not have any test failures. How did you find that difference?

caijun:
I looked at the failures in the Travis CI build. 

https://travis-ci.org/tdhock/animint/builds/71210594

I would like to fix them one by one.

caijun:
@tdhock I am working on the failed test of `test-legends.R` and I am stuck here. The order of legend entries in the third plot is reversed comparing to the version produced by current master branch. The following are the two different versions:

The version produced by this big-chunks branch

http://bl.ocks.org/caijun/raw/eecae7f8b0e96c53ae75/

The version produced by current master branch

http://bl.ocks.org/caijun/raw/6f703085d7c1381b55c3/

Since my modifications are only relevant to how to generate chunk tsv files and make the renderer work with the new chunk tsv file structures. I believe this may be not caused by my modifications. Moreover, I checked the file history of `test-legneds.R`

https://github.com/tdhock/animint/commits/master/tests/testthat/test-legends.R

I think @tdhock you may know the reason why this test is failed.

tdhock:
first you should pull the changes from current master into this branch:

```
git checkout big-chunks
git pull origin master
```

then see if the tests are still failing.

caijun:
@tdhock the `test-legends.R` has been passed. However, the Travis CI build is failed with 

```
The command "Rscript -e "devtools::install(); source('tests/testthat.R', chdir = TRUE)"" exited with 137.
```
Do you know how it happened?

tdhock:
usually when that happens there is some problem with Travis, which may be
fixed by simply restarting the build. I just clicked the restart build
button so let's wait and see what it says after it re-runs the build.

caijun:
The build process has been killed, but I don't know the reason.

cpsievert:
http://stackoverflow.com/questions/27564156/home-travis-build-sh-line-41-pid-killed-exit-code-137

caijun:
@cpsievert I tried to build using container-based infrastructure, but bootstrap needs sudo to execute.

```
The command "./travis-tool.sh bootstrap" failed and exited with 1 during .
```

cpsievert:
@caijun yea, container-based builds aren't feasible for R (yet). I think the only option here is to modify your tests so that they require less memory.

caijun:
@cpsievert as you suggested, I use only two seasons of FluView data to test, which requires less memory. Now the Travis CI build failed with the same error returned on my local PC. 

@tdhock I run `tests_run()` on my local computer, the tests results are as follows

```
testthat results ==================================================================================================
OK: 556 SKIPPED: 0 FAILED: 4
1. Failure (at test-shiny.R#37): animint plot renders in an interactive document 
2. Error: .variable and .value makes compiler create selectors 
3. Failure (at test-variable-value.R#176): some chunks are not downloaded 
4. Failure (at test-variable-value.R#177): some chunks are not downloaded 
``` 

https://github.com/tdhock/animint/commit/82f524180f5940e7413854a34ada8fa1b1b8dbf9

Is the test added by this commit  pushed by you currently passed on your local computer? It failed on my local computer.

tdhock:
Indeed these tests fail on my computer as well!

They test for showSelected/clickSelects.variable/value functionality that I previously committed to this branch, and that @caijun deleted. However the tests were never deleted so they are still failing.

Good thing I still have a copy of the code! I re-merged it into this branch and saw that the ".variable and .value makes compiler create selectors" test now passes.

However there seems to be some problem with the "some chunks are not downloaded" test -- it raises a JavaScript error in the JS code about copying data from the common chunk file. @caijun can you please investigate? Maybe you just have to avoid adding to the number of chunks in the download status table?

Also I saw a new bug in test-widerect.R: the lines are all rendered at opacity=0.6 by default. This is a bug since only the selected lines should be rendered at opacity=0.6 (the others should be rendered at opacity=0.1).

tdhock:
the "line opacity initially 0.1 or 0.6" test passes on my computer, but seems to be failing on travis/PhantomJS. @caijun can you see if it works on your computer on firefox or PhantomJS?

caijun:
@tdhock Sorry I accidentally reset your commits. Could you push them again? I will look at the tests that you added.

tdhock:
I did push my commits back to the big-chunks branch again. As you can see
now the ".variable and .value makes compiler create selectors" test passes
in test-variable-value.R.

However now there seem to be some other test failures related to your new
varied/common chunk code, so could you please investigate?

caijun:
I reset the commits you pushed today. Could you push them once again?

tdhock:
@caijun I merged your changes and pushed these commits again.

caijun:
@tdhock thank you. I am working on the failed tests.

caijun:
the "line opacity initially 0.1 or 0.6" test passed on my computer with firefox, but failed with PhantomJS, which returned following failure.

```
1. Failure (at test-widerect.R#95): line opacity initially 0.1 or 0.6 --------------------------------------------
all(unselected.opacity == 0.1) isn't true
```

The `test-geom-abline.R` passed with both firefox and PhantomJS.

tdhock:
@cpsievert any idea about what to do about this line opacity test which is passing on firefox but not on PhantomJS?

cpsievert:
Not sure, but we already skip some tests in widerect.R on phantomjs, so I guess that's an option. 

(at some point we should really look at doing [a build matrix](http://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) against different browsers)

caijun:
@tdhock could you upload your working example to gist so that I can compare the `plot.json` file content to determine whether in `animint.R` or `animint.js` there is a bug.

tdhock:
all the files are included in the gist https://gist.github.com/tdhock/5e1e0fbcb37e6d2072b1 does that help or did you want to see something else?

tdhock:
however I am seeing some other test failures on firefox:

```
1. Failure (at test-save-separate-chunks.R#191): save separate chunks for non-spatial geoms with repetitive field and multiple vars selected 
4. Failure (at test-WorldBank-good.R#52): bars render 
```

caijun:
All tests in `test-save-separate-chunks.R` have passed on my PC with both PhantomJS and firefox.
```
> tests_run(filter = "save-separate-chunks")
testthat results ==================================================================================================
OK: 30 SKIPPED: 0 FAILED: 0
```

The `Failure (at test-WorldBank-good.R#52): bars render ` hasn't been resolved. I am working on it for a while, but still have no idea how to fix it. Could you help me look at it and give me a hint as I am not very familiar with JS?

caijun:
@tdhock I use current master branch to generate WorldBank-good without time animation viz. When I looked at firebug, there were the same NaN for parsing attributes. Indeed, in `geom6_bar_bar_chunk16.tsv` from either `big-chunks` or `current master` there are some NAs for some columns, for example, `x`, `xmin`, and `xmax`. So I am sure that the NAs occurring in

https://gist.githubusercontent.com/tdhock/ad2b0c923de6b971b9ed/raw/ab5c1529cba6fc4dbdba1f9576654632f90caf4d/geom6_bar_bar_chunk16.tsv

is correct.  Maybe the D3.js can not accept some attributes to be NAs. So  in `download_chunk` function we should add a bit of code to deal with the NAs

https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L732-L755

First we need determine what values those NAs will be converted into.

tdhock:
for NA values we should just delete/ignore those rows in the compiler (no need to save them in the tsv files, that would be a waste of time and disk space)

tests/testthat/test-WorldBank-NA.R
```diff
@@ -0,0 +1,96 @@
...
+common <- read.table(common.tsv, sep="\t", header=TRUE,
+                     comment.char="", quote="")
+
+test_that("common chunk contains expected columns", {
+  expected.cols <-
+    c("ymin", "ymax", "xmin", "fill",
+      "clickSelects", "showSelectedlegendfill",
+      "group")
+  expect_identical(sort(names(common)), sort(expected.cols))
+})
```
tdhock: @caijun can you please modify the R compiler code that you wrote for splitting data into common and varied chunks, so that this test passes?

in this geom_bar the only thing that is variable between subsets is xmax. You should save the group (country) and use it to copy data from the common data whenever a new varied chunk tsv file is downloaded.

tests/testthat/test-WorldBank-NA.R
```diff
@@ -0,0 +1,96 @@
...
+info <- animint2HTML(no.time)
+
+chunk1.tsv <- file.path("animint-htmltest", "geom6_bar_bar_chunk1.tsv")
+chunk1 <- read.table(chunk1.tsv, sep="\t", header=TRUE,
+                     comment.char="", quote="")
+
+test_that("chunk1 contains expected columns", {
+  expect_identical(names(chunk1), c("xmax", "group"))
+})
+
```
tdhock: hey @caijun can you please check your common chunk R compiler code, to see why this test (and the "common chunk contains expected columns" test below) is not passing?

caijun:
@tdhock I am working on it.

caijun:
@tdhock this commit makes the WorldBank with time animation viz works. See following gists

Without time animation

http://bl.ocks.org/caijun/raw/b3c0106123e4203625e3/

With time animation

http://bl.ocks.org/caijun/raw/12e9c06f5d92ff2255ad/

Next I will fix the bug of NAs.

caijun:
@tdhock this commit will remove all NAs when saving chunks. It makes `test-WorldBank-NA.R` passed. The test results of `test_run` with firefox on my Mac is as follows

```
testthat results =================================================================================================
OK: 600 SKIPPED: 0 FAILED: 1
1. Failure (at test-shiny.R#37): animint plot renders in an interactive document
```

After I fixed the bug occurred in `test-shiny.R`, I should be able to merge this branch into master.

tdhock:
You don't need to worry about fixing the test-shiny.R -- that is a false postive test failure.

On travis I see

```
testthat results ================================================================
OK: 585 SKIPPED: 0 FAILED: 3
1. Failure (at test-save-separate-chunks.R#215): save separate chunks for non-spatial geoms with repetitive field, multiple vars selected, and NAs 
2. Failure (at test-variable-value.R#245): changing problem downloads one chunk 
3. Failure (at test-widerect.R#95): line opacity initially 0.1 or 0.6 
```

I don't get any of those errors on firefox on my computer, so I guess those are false positives too. Please only run these tests on firefox, by adding

```
if (Sys.getenv("TRAVIS") == "true") {
  message("tests currently don't work on travis (but should someday)")
} else {
  ## Run the tests!
}
```

@kferris10 @cpsievert can you please code review and run some tests with this branch to make sure everything is working as you would expect?

@caijun you may want to start thinking about writing a blog post that summarizes your GSOC project and the contributions in this PR. For example, "before this PR the FluView data viz takes XXX megabytes, after it takes only XXX megabytes, and so now we can post it on gist..."

R/animint.R
```diff
@@ -402,33 +402,55 @@ saveLayer <- function(l, d, meta){
...
 
-  is.cs <- is.clickSelects(names(g$aes))
+  is.cs <- names(g$aes) %in% s.aes$clickSelects$one
   update.vars <- g$aes[is.ss | is.cs]
 
+  update.var.names <- if(0 < length(update.vars)){
+    data.frame(variable=names(update.vars), value=NA)
+  }else{
+    ##browser()
+  }
```
cpsievert: @caijun please get rid of the else statement here

cpsievert:
@caijun please try to avoid printing objects during tests, for example -> https://travis-ci.org/tdhock/animint/builds/74458410#L1723-L1799

cpsievert:
On my machine on Firefox I see 

```
2. Failure (at test-widerect.R#286): middle of transition != after when duration=1000 
during.width != after.width isn't true

testthat results ================================================================
OK: 599 SKIPPED: 0 FAILED: 2
1. Failure (at test-shiny.R#37): animint plot renders in an interactive document 
2. Failure (at test-widerect.R#286): middle of transition != after when duration=1000 
```

No need to worry about the shiny failure, but I'd like to see that second test pass before we consider merging.

caijun:
@tdhock @cpsievert The following are test results of `test_run()` on my Mac

with PhantomJS

```
testthat results ==================================================================================================
OK: 588 SKIPPED: 0 FAILED: 2
1. Failure (at test-shiny.R#37): animint plot renders in an interactive document 
2. Failure (at test-widerect.R#95): line opacity initially 0.1 or 0.6 
```

with Firefox

```
testthat results ==================================================================================================
OK: 600 SKIPPED: 0 FAILED: 1
1. Failure (at test-shiny.R#37): animint plot renders in an interactive document 
```

cpsievert:
Considering that tests were only 14 minutes as of 3ce29cc, my guess is that one of the tests is stalling phantomjs

caijun:
I removed `test-save-separate-chunks.R` added by myself and the Travis CI build passed with time of about 20 minutes. So I need to optimize this test.

caijun:
@tdhock @cpsievert The Travis CI build passed, however the wercker build failed. Do you have knowledge about the reason why wercker build failed? Next I will write a blog about the optimization of saving disk space for animint. After finishing the blog, I will merge this branch into master. You can review the code and comment on it. Thanks.

cpsievert:
@caijun looking at the build log, I see:

```
Failure (at test-save-separate-chunks.R#220): save separate chunks for non-spatial geoms with repetitive field, multiple vars selected, and NAs 
nrow(varied.data) not equal to 186
186 - 187 == -1
```

Please read the wiki page on [inspecting the wercker build](https://github.com/tdhock/animint/wiki/Testing#inspect-test-builds-locally-with-wercker) and try your best to debug this error.

I also see:

```
Error in time.classes[[1]] : subscript out of bounds
```

So it appears something is going wrong [here](https://github.com/tdhock/animint/blob/f7d47b5a9b019fc3b12c0aa2d822572faf7ce4df/R/animint.R#L1093)

caijun:
@tdhock @cpsievert those commits fix Error in time.classes[[1]] : subscript out of bounds in `test-shiny.R`, referred to https://github.com/tdhock/animint/issues/105

To void stalling PhantomJS with

```
SessionManagerReqHand - _cleanupWindowlessSessions - Asynchronous Sessions clean-up phase starting NOW
```

I have to add TRAVIS check for the test case 1 in `test-save-separate-chunks.R`. Now the TRAVIS CI build passed with 578 tests OK. All tests are passed with Firefox on my local Mac PC. You can run `tests_run()` and report results.

Since the Wercker build failed with the same reason for the same tests, I add WERCKER check to those tests with TRAVIS check. Now the Wercker build failed at the store step. I have no idea how to fix it, but I find that somebody else also encounter the same problem https://github.com/wercker/support/issues/62. @cpsievert do you have any hint to fix this problem?

cpsievert:
Great job @caijun! Don't worry about wercker failing on the store step. I'll try to give this another review today or tomorrow before giving my +1.

caijun:
@tdhock @cpsievert I wrote a blog to summarize my GSOC project and the contributions in this PR and post it to AnimintBlog created by @kferris10 . You can review it and give comments, thanks.

http://caijun.github.io/AnimintBlog/Optimization-of-Disk-Usage-for-animint/

caijun:
@tdhock @cpsievert could you help to review the code? If I can get your +1, then the branch should be able to be merged.

cpsievert:
Sorry for not getting back to you @caijun. I was hoping to find time to do a proper review, but I'm failing to tread water on my other projects. As long as @tdhock is OK with merging, I am too.

PS. nice job on the blog post!

tdhock:
I have pulled changes from current master, and fixed all of the test failure except

```
1. Failure (at test-save-separate-chunks.R#235): save separate chunks for non-spatial geoms with repetitive field, multiple vars selected, and NAs
nrow(varied.data) not equal to 186
186 - 187 == -1
```

which is weird because I don't see that error on phantomjs on my machine. Any ideas?

caijun:
Hi @tdhock I will appreciate your effort to removing some TRAVIS checks that I added when I have no idea how to fix the bugs. Now all tests passed with Firefox on my Mac PC. In my local PC, `varied.data` are really 186 rows. I see the results of TRAVIS build of the `print varied.data` commit, which contains 187 rows.

caijun:
The unique year of WorldBank are 53, while 

```
> all(is.na(WorldBank[WorldBank$year == 2012, "life.expectancy"]))
[1] TRUE
```

After splitting WorldBank by year into chunks, the chunk of 2012 will be empty after removing NAs and will not save to disk. So the number of output chunks is 52, different the number of unique years.

PR #78, opened 2015-05-25
Panel backgrounds

kferris10:
This is a start at coloring background panels.  

In `animint.R`, I [added some code](https://github.com/kferris10/animint/blob/panel_backgrounds/R/animint.R#L64) to extract the `ggplot2:::plot_theme(plot)$panel.background` for each of the panels.

In `animint.js`, to draw the backgrounds I would like to do something [like this](https://github.com/kferris10/animint/blob/panel_backgrounds/inst/htmljs/animint.js#L513)

    svg.append("rect")
          .attr("x", plotdim.xstart)
          .attr("y", plotdim.ystart)
          .attr("width", graph_width)
          .attr("height", graph_height)
         .attr("fill", p_info.panel_background.fill);

However, the result is something like this (I'm trying to give a light blue background to the sepal plot and a light grey background to the petal plot).  There's a problem because it is drawing the rectangles after drawing the points.  

image

If I put that line of code before [this part](https://github.com/kferris10/animint/blob/panel_backgrounds/inst/htmljs/animint.js#L288) then the backgrounds will be drawn before the points.  So I could calculate the locations and sizes of the background rectangles and draw them before this part of the code, but that seems somewhat redundant.  These plot sizes are all calculated in [the for loop](https://github.com/kferris10/animint/blob/panel_backgrounds/inst/htmljs/animint.js#L312) in `add_plot()`.  

Hope all this makes sense.  I'm a bit confused here so I thought I'd check with you guys before continuing on my own.  Thanks!

### DIFF
--- a/R/animint.R
+++ b/R/animint.R
@@ -77,9 +77,9 @@ parsePlot <- function(meta){
       temp_background$linetype <- "solid"
     } else {
       # make sure linetype is correctly specified
-      stopifnot(temp_background$linetype %in% 
-                  c("blank", "solid", "dashed", "dotted", 
-                    "dotdash", "longdash", "twodash"))
+      stopifnot(
+        temp_background$linetype %in% c("blank", "solid", "dashed", "dotted", "dotdash", "longdash", "twodash") | 
+          is.numeric(temp_background$linetype))
     }
   }
   # saving background info
--- a/R/animint.R
+++ b/R/animint.R
@@ -59,6 +59,31 @@ parsePlot <- function(meta){
   ## grid 0-1 scale). This allows transformations to be used
   ## out of the box, with no additional d3 coding.
   theme.pars <- ggplot2:::plot_theme(meta$plot)
+  
+  ## extract pnale backgrounds from theme.pars
+  temp_background <- theme.pars$panel.background
+  # convert fill to RGB if necessary
+  if(!(is.rgb(temp_background$fill))) { 
+    temp_background$fill <- toRGB(temp_background$fill)
+  }
+  # if border color is specified
+  if(!is.na(temp_background$colour)) {
+    # convert color to RGB if necessary
+    if(!(is.rgb(temp_background$colour))) { 
+      temp_background$colour <- toRGB(temp_background$colour)
+    }
+    # check if the user specified linetype for the border
+    if(is.null(temp_background$linetype)) {
+      temp_background$linetype <- "solid"
+    } else {
+      # make sure linetype is correctly specified
+      stopifnot(temp_background$linetype %in% 
+                  c("blank", "solid", "dashed", "dotted", 
+                    "dotdash", "longdash", "twodash"))
+    }
+  }
+  # saving background info
+  plot.meta$panel_background <- temp_background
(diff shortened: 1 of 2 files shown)


R/animint.R
```diff
@@ -59,6 +59,31 @@ parsePlot <- function(meta){
...
+      temp_background$colour <- toRGB(temp_background$colour)
+    }
+    # check if the user specified linetype for the border
+    if(is.null(temp_background$linetype)) {
+      temp_background$linetype <- "solid"
+    } else {
+      # make sure linetype is correctly specified
+      stopifnot(temp_background$linetype %in% 
+                  c("blank", "solid", "dashed", "dotted", 
+                    "dotdash", "longdash", "twodash"))
```
tdhock: why wouldn't integer linetypes be acceptable? they are acceptable in the animint.js renderer, https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L11

inst/htmljs/animint.js
```diff
@@ -510,6 +510,14 @@ var animint = function (to_select, json_file) {
 
     Plots[p_name].scales = scales;
 
+    // draw background
+    svg.append("rect")
+      .attr("x", plotdim.xstart)
+      .attr("y", plotdim.ystart)
+      .attr("width", graph_width)
+      .attr("height", graph_height)
+      .attr("fill", p_info.panel_background.fill);
```
tdhock: you should use `.style` rather than `.attr` to set "fill" "stroke" "stroke-dasharray" and "stroke-width", as in the definition of geoms that make a `<rect>` https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L1093-L1096

tdhock:
this looks like a good start.

to make sure the background rect appears behind the geoms, put it in the `<svg>` element, but before the grouping `<g>` elements (in which the geoms are drawn, https://github.com/kferris10/animint/blob/panel_backgrounds/inst/htmljs/animint.js#L736).

To accomplish that I suggest you use `insert` https://github.com/mbostock/d3/wiki/Selections#insert rather than `append`

kferris10:
[Updated example](http://bl.ocks.org/kferris10/raw/f4e5526c39eec1766a86/)

I think this might take care of most of the problems that you mentioned.  I had to change [`linetypesize2dasharry`](https://github.com/kferris10/animint/blob/panel_backgrounds/inst/htmljs/animint.js#L11)

cpsievert:
Awesome! How hard do you think it'd be to implement `panel.grid.major` and `panel.grid.minor` as well?

kferris10:
I don't think that it should be that hard.  I was going to ask, should this be part of the same PR or should it be a separate one?

cpsievert:
I don't have a strong preference either way, but if you think it won't be too bad, might as well try it out in this pull request (also, please try to write some tests for this new functionality)

cpsievert:
Also, in the future, it'd be great if you could send pull requests from the tdhock remote rather than your fork of tdhock

kferris10:
Moved to #84

Issue #66, opened 2015-05-14
Error running tests_init()


kferris10:
I'm trying to familiarize myself with the testing workflow for animint.  @cpsievert has pointed me to [the testing wiki](https://github.com/tdhock/animint/wiki/Testing) to get started, but I have been unable to run `tests_init()` without an error.

    > library(animint)
    > tests_init()
    
    Starting phantomjs binary. To shut it down, run: 
    pJS$stop()
    Error: 	 Summary: NoSuchElement
     	 Detail: An element could not be located on the page using the given search parameters.
     	 class: org.openqa.selenium.NoSuchElementException

It also spits out some warning messages which Carson said not to worry about.  If I try using Firefox, I get a different error:

    > tests_init("firefox")
    Undefined error in RCurl call.

    Starting selenium binary. To shut it down, run: 
    remDr$closeWindow() 
    remDr$closeServer()
    Undefined error in RCurl call.
    Error in queryRD(paste0(serverURL, "/session"), "POST", qdata = toJSON(serverOpts)) : 

Here's my session info, let me know what else I can do to help!

    > sessionInfo()
    R version 3.1.3 (2015-03-09)
    Platform: x86_64-w64-mingw32/x64 (64-bit)
    Running under: Windows 8 x64 (build 9200)

    locale:
    [1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
    [3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
    [5] LC_TIME=English_United States.1252    

    attached base packages:
    [1] stats     graphics  grDevices utils     datasets  methods   base     

    other attached packages:
    [1] animint_2015.05.11 proto_0.3-10       ggplot2_1.0.0.99  
    
    loaded via a namespace (and not attached):
     [1] bitops_1.0-6     caTools_1.17.1   colorspace_1.2-4 digest_0.6.8    
     [5] grid_3.1.3       gtable_0.1.2     MASS_7.3-39      munsell_0.4.2   
     [9] plyr_1.8.1       Rcpp_0.11.5      RCurl_1.95-4.5   reshape2_1.4.1  
    [13] RJSONIO_1.3-0    RSelenium_1.3.5  scales_0.2.4     stringr_0.6.2   
    [17] tools_3.1.3      XML_3.98-1.1

cpsievert:
First, lets make sure minimal examples with **RSelenium** work:

* phantomjs:

```r
[...]

PR #79, opened 2015-05-25
Fix ":262 in error" when running tests_init()

caijun:
As reported in tdhock/animint#66, this branch fixes the `:262 in error` when running `tests_init()` after `tests_init("firefox)`. My idea is to get process ids by port number used in animint. `tests_init()` should run in the environment that no processes occupy the port number. Also, this makes the code more succinct comparing by using `pid.txt`.

The branch works well in creating and exiting remote server of selenium on my Mac OS X 10.10.3, even though the following error and warning occur.

```
Error in startServer(host, port, app) : Failed to create server
Calls: <Anonymous> -> <Anonymous> -> startServer
Execution halted
```
This error is likely to result from `httpuv::startServer()`, which I haven't trace back.
```
Warning message:
running command 'lsof -i :4444 -t' had status 1 
```
This warning has no influence to our functions. Hence, I will suppress it in the future.

@tdhock and @cpsievert can review the changes and give comments.

### DIFF
--- a/R/testHelpers.R
+++ b/R/testHelpers.R
@@ -99,7 +99,6 @@ tests_run <- function(dir = ".", filter = NULL) {
 #' @seealso \link{tests_run}
 #' @export
 tests_exit <- function() {
-  
   res <- stop_binary()
   Sys.unsetenv("ANIMINT_BROWSER")
   # release port used for local file server
--- a/R/testHelpers.R
+++ b/R/testHelpers.R
@@ -99,14 +99,14 @@ tests_run <- function(dir = ".", filter = NULL) {
 #' @seealso \link{tests_run}
 #' @export
 tests_exit <- function() {
+  
   res <- stop_binary()
   Sys.unsetenv("ANIMINT_BROWSER")
-  e <- try(readLines(pid_file(), warn = FALSE), silent = TRUE)
-  if (!inherits(e, "try-error")) {
-    pids <- as.integer(e)
-    res <- c(res, tools::pskill(pids))
+  # release port used for local file server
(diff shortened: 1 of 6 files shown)


tdhock:
I tried running the tests and they paused/stopped during the test-facet-trivial.R tests.

R/testHelpers.R
```diff
@@ -164,3 +160,12 @@ find_test_path <- function(dir = ".") {
...
   file.path(dir, ext_dir)
 }
+
+# get process id via port number
+port2pid <- function(port = 4444) {
+#   args <- paste0("-i :", port, " -t")
+#   pid <- as.integer(system2("lsof", args = args, stdout = TRUE))
+  cmd <- paste0("lsof -i :", port, " -t")
+  pid <- as.integer(system(cmd, intern = TRUE))
+  pid
```
cpsievert: this won't work cross-platform

cpsievert:
I looked into killing process(es) by port number a while ago, but I couldn't find a nice cross-platform way to do that. 

I think you're problems in #66 are due to selenium not shutting down properly (which might be a problem with RSelenium internals) so let's make sure we identify the issue before spending anymore time here.

cpsievert:
I reopen this if we can't sort out #66 without it, but I'm pretty this isn't the source of our problems...

R/testHelpers.R
caijun: Yes, it is a solution for Unix-like OS, not Windows.

caijun:
I update #66 , it is a problem of  `RSelenium::closeServer()`, which doesn't kill `java` process properly.

PR #81, opened 2015-05-27
legend entry order

tdhock:
This branch contains a test that fails for the order of legend entries.

First of all we need to attach the "legend_entry_label" class to the legend entry label `<td>` elements.

Then we need to fix the compiler so that it outputs the continuous legend entries in the correct order.


tdhock:
For numeric legends the smaller values are now placed lower in the legend, as in http://bl.ocks.org/tdhock/raw/ffe5fe7f693564aba165/

Tests pass on firefox on my machine. This PR is ready to merge, can I get a +1 please @cpsievert?

cpsievert:
:+1:

PR #84, opened 2015-06-01
Panel backgrounds

kferris10:
Moved from #78 

In addition to drawing background rectangles, I amended [the linetype test](https://github.com/tdhock/animint/blob/panel_backgrounds/tests/testthat/test-linetype.R#L19) because the background rectangles were interfering with this test.  If somebody could confirm that this makes sense, that would be great.

I've drawn the [major grid lines](https://github.com/tdhock/animint/blob/panel_backgrounds/inst/htmljs/animint.js#L527) and will get to the minor ones as soon as possible


tests/testthat/test-linetype.R
```diff
@@ -15,11 +16,10 @@ dasharrayPattern <-
          ";")
 
 rect.xpaths <- 
-  c('//svg[@id="numeric"]//rect',
-    '//svg[@id="character"]//rect',
+  c('//svg[@id="numeric"]//g[@class="PANEL1"]//rect',
+    '//svg[@id="character"]//g[@class="PANEL1"]//rect',
     '//td[@id="numeric_legend"]//rect',
```
tdhock: Hey Kevin thanks for the nice description of your modifications. Yes it does make sense to change this test. Next time please consider using an inline comment like this so we can also see the diff.

tdhock:
also do you think it would be easy to support `theme(panel.margin=grid::unit(0, "cm"))`?
I often use it with `theme_bw()` to eliminate the space between panels.

R/animint.R
```diff
@@ -59,6 +59,63 @@ parsePlot <- function(meta){
...
+              is.numeric(pars$linetype))
+        }
+      }
+    }
+    pars
+  }
+
+  # saving background info
+  plot.meta$panel_background <- get_bg(theme.pars$panel.background)
+  plot.meta$panel_border <- get_bg(theme.pars$panel.border)
```
kferris10: Here, I'm extracting the `panel_background` and `panel_border` separately.  I believe that `panel_background` is used to `fill` the background rectangle and `panel_border` is used to draw the border around each panel

kferris10:
@tdhock I think that should be possible.  I'll take a deeper look at it tomorrow, but I think that I should be able to adjust the plot margins to make it work.

I just realized there's a bug that's causing the tests to fail, but other than that I think that these updates should take care of the backgrounds.  We can now shade the backgrounds different colors, draw different borders, and draw both the major and minor grid lines.

I want to ask one question about this process.  If you look at the inline comment above (hope I did that right), I'm pretty sure ggplot2 uses `panel.background` to shade in the backgrounds and `panel.border` to draw the borders.  Does anybody know if that's the right idea?

tdhock:
I think you are right about panel.background and panel.border:

```
> theme_bw()$panel.background
List of 4
 $ fill    : chr "white"
 $ colour  : logi NA
 $ size    : NULL
 $ linetype: NULL
 - attr(*, "class")= chr [1:2] "element_rect" "element"
> theme_grey()$panel.background
List of 4
 $ fill    : chr "grey90"
 $ colour  : logi NA
 $ size    : NULL
 $ linetype: NULL
 - attr(*, "class")= chr [1:2] "element_rect" "element"
> theme_bw()$panel.border
List of 4
 $ fill    : logi NA
 $ colour  : chr "grey50"
 $ size    : NULL
 $ linetype: NULL
 - attr(*, "class")= chr [1:2] "element_rect" "element"
> theme_grey()$panel.border
 list()
 - attr(*, "class")= chr [1:2] "element_blank" "element"
```

to make sure, you can always print the ggplots in R and look at them.

kferris10:
Thanks. I think I've got this about ready, I just have two questions.

1. In animint.js, you initially specify the margin for each panel as
`margin: {top:0, bottom:10, left:0, right:10}`. Is there a reason you do this instead of `margin: {top:5, bottom:5, left:5, right:5}`?

2. In `theme(panel.margin = ...)` can be specified using a bunch of different units. Does anyone know of a function I could use to convert all these units to pixels? It would help a bit

cpsievert:
As for (1), I think the right space is for long tick labels, but not too sure about the bottom (maybe cushion for axis name text?). As for (2), let's start with supporting "npc". I think you'll have to write a custom translator that accounts for the specified height/width (of `theme_animint()`)

kferris10:
For (1), I believe that the margin is controlling the spacing between panels. I thought that tick labels were handled inside the panel itself. Is that incorrect?

For (2), do you think that accounting for theme_animint is necessary? We could just let people try out a few different margins until they find one they like

cpsievert:
I don't think we adjust for the width of tick labels. So if right-most x-axis tick label is super long, the text would run-off, but 10 seems to work for most cases.

I wouldn't say it's _necessary_, but it'd be nice ;)

tdhock:
animint.js definitely does adjust the size of the bottom border based on the text on the x axis ticks, see https://github.com/tdhock/animint/blob/master/tests/testthat/test-rotate.R for some examples.

kferris10:
Good point @cpsievert .  I'll see what I can do.

I'm pushing the updates so you guys can look at it, but they aren't ready to be merged yet.  I still don't have everything working quite right when using bars or other discrete variables (I think that's the reason why so many tests are failing).  I'll try to get that fixed shortly.

I also [added an example](https://github.com/tdhock/animint/blob/panel_backgrounds/inst/examples/panels.R) demonstrating different things you can do.  You can view a [live example here](http://bl.ocks.org/kferris10/raw/10d628b74c3d69fa9d55/)

kferris10:
The previously mentioned issue is now fixed.  Turns out, it had nothing to do with bars and was an issue if grid lines weren't supplied.  All the tests now pass for me

I've also started some tests for these different panels.  I'm running into an issue, however, and I'm hoping somebody can point me in the right direction.

In `test-panels.R`, I've inserted a `browser()` call. https://github.com/tdhock/animint/blob/2fd5e6aa85073700c5bb545ec03c8c47124c400a/tests/testthat/test-panels.R#L75  If you uncomment this, and type `attr_back_petal` I get this.

           [,1]                       
    x      "37"                       
    y      "33.25"                    
    width  "357.75"                   
    height "305.5"                    
    class  "background_rect"          
    style  "fill: rgb(229, 229, 229);"

But when I do `attr_border_petal` the style information is not extracted.  Does anybody know why this is?

           [,1]         
    x      "37"         
    y      "33.25"      
    width  "357.75"     
    height "305.5"      
    class  "border_rect"

tdhock:
Using firefox on Ubuntu I get style attributes in both cases

```
> attr_border_petal
       [,1]                                            
x      "42.25"                                         
y      "34.25"                                         
width  "352.5"                                         
height "302.5"                                         
class  "border_rect"                                   
style  "fill: transparent; stroke: rgb(127, 127, 127);"
> attr_back_petal
       [,1]                       
x      "42.25"                    
y      "34.25"                    
width  "352.5"                    
height "302.5"                    
class  "background_rect"          
style  "fill: rgb(255, 255, 255);"
> 
```

kferris10:
Weird, could this just be a Windows issue?

cpsievert:
On Firefox, I'm also seeing that the style attribute is empty for `attr_border_petal`.

screen shot 2015-06-09 at 10 17 02 am

 Is it supposed to be transparent? Maybe, in certain cases, this behavior/result is to be expected?

kferris10:
I messed this up by forgetting to update my example in "test-panels".  My apologies.

I think that the wercker tests are failing because it is checking hexadecimal color values.  I'm using the rgb colors because that is what is generated by `animint2dir`

image

Should I switch to checking hexadecimal colors?

tdhock:
here is an example of how I tested colors: https://github.com/tdhock/animint/blob/master/tests/testthat/test-params.R#L48-L55

kferris10:
Got it.  I think this takes care of it.  

I haven't started working on the panel margins yet.  I think that will take some work.  Would it make sense to merge this branch now and create a new PR for the margins?

tdhock:
Thanks for contributing this great code and tests Kevin.

Yes it makes sense to merge this branch without implementing the panel.margin feature.

However I ran the test suite and saw the following visual regression (in both firefox and chrome)

facet_top_cutoff

The top facet labels are cutoff (hard to see in "Years" but the top of the F and the dot on the i are clearly missing in "Fertility rate").

Can you please add a test that fails in test-widerect.R?

And then can you please fix that bug?

kferris10:
Thanks for catching that.  It originated a couple of days ago when I was working on panel margins.  I've removed it, 

I'd prefer to add the test when I start working on the panel margins.  Because that is what cause the issue, I think it'd be more consistent to put it there.

R/animint.R
```diff
@@ -991,7 +1047,11 @@ servr::httd("', normalizePath( out.dir,winslash="/" ), '")')
 #' @return True/False value
 #' @export
 is.rgb <- function(x){
-  grepl("NULL", x) | (grepl("#", x) & nchar(x)==7)
+  if(is.null(x)) {
+    TRUE
+  } else {
+    (grepl("#", x) & nchar(x)==7)
```
cpsievert: Maybe we could change this line to `grepl("^#[0-9A-Za-z]{6}$", x)` ?

cc @tdhock

caijun: +1

cpsievert:
@kferris10 can't wait to merge this (looks great)!! It's a huge improvement especially for facetted plots that share an axis.

I found it a bit surprising that minor grid lines are drawn over major ones. Is that intentional (is that how ggplot2 does it)? I wrote a test to check the number of minor grid lines, but feel free to modify it if that behavior is to be expected.

kferris10:
It looks like that's just how ggplot2 does it.  Check out the major vs minor

```
str(ggplot_build(p2)$panel$ranges[[1]])
List of 12
 $ x.range       : num [1:2] 0.705 7.195
 $ x.labels      : chr [1:3] "2" "4" "6"
 $ x.major       : num [1:3] 0.2 0.508 0.816
 $ x.minor       : num [1:7] 0.0455 0.1995 0.3536 0.5077 0.6618 ...
 $ x.major_source: num [1:3] 2 4 6
 $ x.minor_source: num [1:7] 1 2 3 4 5 6 7
 $ y.range       : num [1:2] -0.02 2.62
 $ y.labels      : chr [1:6] "0.0" "0.5" "1.0" "1.5" ...
 $ y.major       : num [1:6] 0.00758 0.19697 0.38636 0.57576 0.76515 ...
 $ y.minor       : num [1:11] 0.00758 0.10227 0.19697 0.29167 0.38636 ...
 $ y.major_source: num [1:6] 0 0.5 1 1.5 2 2.5
 $ y.minor_source: num [1:11] 0 0.25 0.5 0.75 1 1.25 1.5 1.75 2 2.25 ...
```

I thought that I was drawing the major lines on top of the minor ones.  Is that correct?

tdhock:
I tried implementing a test for the unreadable strip labels but it seems complicated. How to tell if the svg `<text>` element is partly overlapping the top of the `<svg>` element?

I tried adding the code below to test-facet-trivial.R:

```r
test_that("top strip labels are not cut off", {
  viz <-
    list(Sk=gg+facet_grid(Species ~ kingdom))
  info <- animint2HTML(viz)
  strip.rect <- remDr$executeScript('return document.getElementById("topStrip").firstChild.getBoundingClientRect()')
  svg.rect <- remDr$executeScript('return document.getElementById("Sk").getBoundingClientRect()')
})
```

cpsievert:
@kferris10 I'm not positive, but from these examples, it appears ggplot2 doesn't draw the minor grid lines where the major ones are (I think prefer that behavior anyway so we don't clutter up the DOM)

```r
qplot(Sepal.Length, Sepal.Width, data = iris) + theme( panel.grid.minor = element_line("blue"))
```

rplot

```r
qplot(Sepal.Length, Sepal.Width, data = iris) + theme(panel.grid.major = element_line("red"), panel.grid.minor = element_line("blue"))
```

rplot01

tdhock:
I agree with Carson. The less DOM elements to create, the faster the
animint viz will be. (browsers are known to slow down after adding more
than 10,000 elements to an SVG element).

kferris10:
@tdhock What about this instead?  For this check, we just need to make sure that the text is pushed down more than the font size, right?

```
fontsizePattern <- 
  paste0("font-size: ", 
         "(?<value>.*?)", 
         "px;")
```
```
  top_text <- getNodeSet(info$html, '//g[@id="topStrip"]//text')
  top_attrs <- sapply(top_text, xmlAttrs)
  top_translate <- str_match_perl(top_attrs["transform",], translatePattern)
  top_font_size <- str_match_perl(top_attrs["style",], fontsizePattern)
  expect_true(as.numeric(top_translate[3]) > as.numeric(top_font_size[2]))
```

tdhock:
in theory your argument makes sense Kevin.

However in practice on my firefox browser the default translate value of 7.25 cuts of the 11px text:

strip-translate-7 25-default

but a value of 8 (less than 11px) is sufficient for a strip label that is not cut off:

strip-translate-8

or am I missing something?

in cases like this where the test is not straightforward to write, I prefer to just not write a test (avoid false test failures in the future).

tdhock:
tests pass on my computer with firefox. +1 go ahead and merge when you like.

cpsievert:
Great, this also has my +1, make sure you update the version in the DESCRIPTION and the NEWS before (or after) merging @kferris10. 

It'd also be awesome if you could re-build the tutorial to reflect the new theming defaults!

kferris10:
Not a problem.  Are there any other visualizations that should be rebuilt?

Also, should I just bump the version to today's date?

cpsievert:
Just the tutorial is fine for now. And yea, use today's date.

kferris10:
Great, this last commit should take care of that.  

This means that we can also merge #68 as well, right?

cpsievert:
Cool, go ahead and merge this.

After you merge this into master, you might have conflicts when you pull these changes into #68 (waiting to update the version and NEWS helps in this regard). Are you familiar with resolving merge conflicts? If it is a big conflict, sometime I do `git mergetool` which pops up a helpful GUI for resolving the conflict.

kferris10:
Just found a bug when compiling the animint tutorial.  Does it make the most sense to push the fix to this branch or to create a new one?

tdhock:
can you please first write a test that fails for the viz with the bug?

after that I would suggest pushing the changes to a new branch and making a new PR.

kferris10:
Already done :)

kferris10:
[I've updated the tutorial](http://kferris10.github.io/animint/).  @srvanderplas could you take a quick look at it when you get a chance?  Three things in particular:

1. I set the climate visualization to remove grid lines, set a white background, and added to your explanation of the `theme`.
2. For some reason, there's now a problem with `scale_y_` in the density examples.  I commented out this line in the [geom_polygon with stat = 'density2d' contours](http://kferris10.github.io/animint/#geom_polygon-with-stat-density2d-contours) example but left it in the [scale_y_log10() and geom_contour() with stat_density2d()](http://kferris10.github.io/animint/#scale_y_log10-and-geom_contour-with-stat_density2d) example so you can see what I'm talking about.  Figured I'd see if you have any idea what's going on before I dig into it.
3. In the `abline` example, I don't see the lines generated in either [your example](http://tdhock.github.io/animint/#geom_abline) or in [the updated version](http://kferris10.github.io/animint/#geom_abline).  Any idea what is going on here?

tdhock:
for `geom_abline` there are `<line>` elements with `NaN` values.

abline-problem

Not sure when this bug was introduced, but it did work at some point in the past.

Issue #71, opened 2015-05-20
Add test coverage reports for animint


caijun:
I think it's better to add test coverage reports for animint. The [covr](https://github.com/jimhester/covr) package can accomplish such a task.

tdhock:
What exactly is the benefit of test coverage reports? If you think the benefit is great enough, then go ahead and implement it for animint.

I guess the covr package will compute something based on animint's R code, but will ignore the JS code, right? Is that a problem?

caijun:
Test coverage reports describe the degree to which the source code of animint is tested by the tests. It help to design efficient and effective test cases, using less test cases to test as more codes as possible. The reports will show which parts of source code haven't been tested and test cases need to design to avoid bugs.

I will implement it for anmint and to see whether or not the JS code will be ignored.

tdhock:
It definitely sounds like a good idea in theory to test the "degree to which the source code of animint is tested by the tests" and to use "less test cases" to test as much code as possible, but I suppose it may be complicated to compute something that is meaningful for us based on our R+JS code base. Anyways go ahead and try it -- maybe having just the default covr analysis of our R code is better than nothing.

tdhock:
moving to #85

PR #85, opened 2015-06-01
add test coverage with coveralls

caijun:
This `covr` branch adds test coverage for animint. I have turned on coveralls for animint and the Travis build is passed. After `covr` merged, there will be a coverage status badge in README.md. The current test coverage is `77%`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,13 @@ install:
   - $HOME/opt/pandoc/pandoc --version
   # install uninstalled packages which this pkg depends on/links to/suggests. 
   - ./travis-tool.sh install_deps
-  - ./travis-tool.sh install_github tdhock/ggplot2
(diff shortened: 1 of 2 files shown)


tdhock:
As discussed in #71 this 77% refers to only the R code right?

There are many libraries for computing coverage for JS code. For example there is http://blanketjs.org/ but that would not work for us since it requires a JS testing framework (Qunit, Mocha or Jasmine). I suppose that our testing of animint.js by writing test-*.R files is rather unique in the open-source world so I doubt that there is an existing library that will easily let us compute our JS code coverage. So I guess we should just forget about trying to compute JS code coverage at least for now.

Any other ideas?

caijun:
I also think 77% is only the test coverage for R code and JS code is not included. I have set the test coverage aside and moved on to the optimization of `saveChunks` function for my FluView example. Prioritizing the tasks listed in my proposal will make a successful google summer of code.

PR #86, opened 2015-06-17
Hotfixes

kferris10:
Fix bug which occurs when there were no minor grid lines for either the x or y axis


kferris10:
Can I get a +1 before merging?  

For the sex/smoker example in the tutorial, there are no minor grid lines on either the x or y axis.  Previously, this would break some of the code I'd written to draw grid lines because I wouldn't be able to access the `grid_minor.loc` object.  There may be a better way to do this, but I think that this works

cpsievert:
Nice, +1!

PR #87, opened 2015-06-19
Hotfixes

kferris10:
Realized that I didn't handle `element_blank()` correctly for the panel backgrounds.  This should take care of that, and I've added a test as well.  

animint now can handle the themes in the [ggthemr](https://github.com/cttobin/ggthemr) and most of the themes in the [ggthemes](https://github.com/jrnold/ggthemes) package.  There are a few themes in `ggthemes` that don't work.  I'll figure this out soon.

### DIFF
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -512,76 +512,82 @@ var animint = function (to_select, json_file) {
       
       // drawing border
       // uses insert to draw it right before the #plottitle
-      svg.insert("rect", "#plottitle")
-        .attr("x", plotdim.xstart)
-        .attr("y", plotdim.ystart)
-        .attr("width", plotdim.xend - plotdim.xstart)
-        .attr("height", plotdim.yend - plotdim.ystart)
-        .attr("class", "border_rect")
-        .style("fill", p_info.panel_border.fill)
-        .style("stroke", p_info.panel_border.colour)
-        .style("stroke-dasharray", function() {
-          return linetypesize2dasharray(p_info.panel_border.linetype,
-                                        p_info.panel_border.size);
-        });
+      if(Object.keys(p_info.panel_border).length > 1) {
+        svg.insert("rect", "#plottitle")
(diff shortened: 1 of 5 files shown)


inst/htmljs/animint.js
```diff
@@ -512,76 +512,82 @@ var animint = function (to_select, json_file) {
...
+                .attr("y2", function() { return scales[panel_i].y(element); })
+                .attr("class", function() { return "grid " + grid_class; })
+                .style("stroke", col)
+                .style("stroke-linecap", cap)
+                .style("stroke-width", size)
+                .style("stroke-dasharray", function() {
+                  return linetypesize2dasharray(lt, size);
+                });
+            }
+            grid_background.loc.y.forEach(draw_hor_line);
```
tdhock: To add grid lines you used forEach, which is fine, but I would have used a d3.js data-bind. The advantage of using a data-bind is that the elements would be easier to update. However that doesn't really matter for grid lines since they are added once at the beginning of plot rendering and then they do not change. Anyways if you are interested in learning about d3 data-binds I would suggest reading https://github.com/mbostock/d3/wiki/Selections#d3_selectAll and http://bost.ocks.org/mike/circles/

cpsievert:
Awesome, +1!

tdhock:
"There are a few themes in ggthemes that don't work" -- it would be nice if you added a test for the theme that you saw which did not work.

However I don't think this is a big deal so you can go ahead and merge when you like +1.

kferris10:
Does animint support legends that are beneath the plot?  If it doesn't, that might be the problem.

tdhock:
animint does not support legend positions other than on the right!

kferris10:
That actually wasn't the problem.  I'm not really sure what's causing the problem, and I'd rather not spend my time on other stuff so I'll just go ahead and merge this if that's okay.

tdhock:
yeah its fine to merge, but could you at least post an issue with the problem that you found?

PR #90, opened 2015-06-24
Hotfixes

kferris10:
I was trying to figure out why the plot in the tutorial with a log transformed y-axis weren't working, but I found a different problem instead.  If there was only one grid line being drawn, I had to coerce the object to an array.  (See the inline comment)

I also took this as an opportunity to clean up the background, grid lines, and borders by nesting them in g elements and to switch to data-binds to draw the grid lines instead of forEach loops.

### DIFF
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -531,50 +531,64 @@ var animint = function (to_select, json_file) {
           });
       }
       
-      // function to draw the grid lines when supplied with one of the elements
+      // function to draw major/minor grid lines 
       var grid_line = function(grid_background, grid_class) {
+        // if grid lines are defined
         if(Object.keys(grid_background).length > 1) {
           var col = grid_background.colour;
           var lt = grid_background.linetype;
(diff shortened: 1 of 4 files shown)


inst/htmljs/animint.js
```diff
@@ -543,56 +531,87 @@ var animint = function (to_select, json_file) {
...
-                .style("stroke", col)
-                .style("stroke-linecap", cap)
-                .style("stroke-width", size)
-                .style("stroke-dasharray", function() {
-                  return linetypesize2dasharray(lt, size);
-                });
-            }
-            grid_background.loc.y.forEach(draw_hor_line);
+            // coercing y lines to array if necessary
+            if(typeof grid_background.loc.y == "number") grid_background.loc.y = [grid_background.loc.y];
```
kferris10: If it's a number, coerce it to an array.  If somebody knows of a better way to do this, let me know.

cpsievert:
Nice work, +1.

PR #91, opened 2015-06-26
Automatically re-build tutorial when pushing to master branch

kferris10:
@cpsievert Here's my attempt at telling travis to automatically re-build the tutorial.  Let me know how it looks (because I'm not entirely sure if what I'm trying to do is right).


.travis.yml
```diff
@@ -3,11 +3,12 @@ language: c
 before_install:
   - curl -OL http://raw.github.com/craigcitro/r-travis/master/scripts/travis-tool.sh
   - chmod 755 ./travis-tool.sh
+  - chmod 755 ./.push_gh_pages.sh
```
kferris10: Like this?

Issue #92, opened 2015-06-29
geom_abline doesn't work


cpsievert:
As [noted here](https://github.com/tdhock/animint/pull/84#issuecomment-112899816), `geom_abline()` isn't working. Here is a mini-example:

```r
library(animint)
p <- qplot(wt, mpg, data = mtcars) + geom_abline(intercept = 20)
animint2dir(list(p = p))
```

kferris10:
Fixed in #93

kferris10:
Re-opening.  Only works for one abline

PR #93, opened 2015-07-03
Fix bug with Geom abline

kferris10:
I think that this should take care of #92 .  The issue was that `ranges` is a list (with one element for each panel), but it was not being treated as such.


cpsievert:
Not sure why the "storage" step of the wercker build is failing, but tests pass, so +1!

PR #96, opened 2015-07-07
Selector widgets

kferris10:
Here's a start to creating a widget for each selector variable.  It's a work in progress as it only works for this one example, but  I wanted to show you guys where I'm currently at.

```
p1 <- ggplot() +
  geom_point(aes(Sepal.Length, Sepal.Width, clickSelects = Species),
             data = iris) +
  ggtitle("Sepal Data")
p2 <- ggplot() +
  geom_point(aes(Petal.Length, Petal.Width, colour = Species, 
                 showSelected = Species), data = iris) +
  ggtitle("Petal Data")

viz <- list(sepal = p1, petal = p2)
animint2dir(viz)
```

It also works with multiple selection

```
viz <- list(sepal = p1, petal = p2, 
            selector.types = list(Species = "multiple"))
animint2dir(viz)
```

[Live example](http://bl.ocks.org/kferris10/raw/47986dcff139b89d6ef7/)


inst/htmljs/animint.js
```diff
@@ -1744,6 +1751,55 @@ var animint = function (to_select, json_file) {
...
+            this.innerHTML = "Hide " + this.id + " selectors";
+          } else {
+            // hide table
+            d3.selectAll(".selector_table").select("#" + this.id)
+              .style("display", "none");
+            // update widget text
+            this.innerHTML = "Show " + this.id + " selectors";
+          }
+        })
+      ;
```
kferris10: When clicking the "Show Species selector" button, I want to update the `selector_table` so that it is no longer hidden.  But this approach isn't working.  I think that there's a problem with the `d3.selectAll(".selector_table").select("#" + this.id)` statement, but I'm not sure what the problem is.  If anybody has any ideas, please let me know.

tdhock:
would it be possible to use http://brianreavis.github.io/selectize.js/ ?

the autocomplete menu would be especially useful for selecting from a long list of things (for example countries in the worldbank viz).

kferris10:
I'm sure it is, but I've never used that before so it might take me a bit.  To start, I'd have to add `selectize.js` and `selectize.css` to `/Animint/inst/htmljs/vendor/` right?  Or is there a more appropriate location?

tdhock:
yes that is an appropriate place to store the selectize css and js files. github's repos code analysis ignores files under the /vendor subdirectory (so then the percent of JS and CSS code in animint is not artificially elevated).

kferris10:
@tdhock Is this sort of how you picturing the widgets visually?  (This is still very much a work in progress).
http://bl.ocks.org/kferris10/raw/853a6222fbde8947e348/

tdhock:
yeah I think that is a good start! Also I think the widget you render should be a function of the selection variable type (single or multiple).

kferris10:
I'm not exactly sure what you are suggesting.  Do you mean that the user should be able to enter multiple levels of a variable, then click an "Update" button and all the selected levels will update simultaneously?

cpsievert:
This shiny example is a good example of how envision this working (selectize examples are on the bottom row) -> http://shiny.rstudio.com/gallery/selectize-vs-select.html

cpsievert:
I don't think an update button is necessary. Just have the input value(s) always reflect the current selection(s)

tdhock:
I agree with Carson, no update button should be necessary, and the menus should work like the last two examples on that shiny page (depending on whether the selection variable is declared as single or multiple)

kferris10:
Gotcha, I see what you're thinking now.  That, plus StackOverflow helping me to understand selectize :), makes much more sense than what I was thinking.  

[Here's a current example](http://bl.ocks.org/kferris10/raw/4af5c85e192d35e49090/)  with selectize working.  As usual, it only works for this example and there's still some work to be done (extending it beyond this example, fixing the size of the button off the top of my head, and fixing for multiple selection).  But I thought you guys might like to see where it's at.

Here's the code I used for the visualization

```
p1 <- ggplot() +
  geom_point(aes(Sepal.Length, Sepal.Width, clickSelects = Species, colour = Species),
             data = iris) +
  ggtitle("Sepal Data")
p2 <- ggplot() +
  geom_point(aes(Petal.Length, Petal.Width, colour = Species, 
                 showSelected = Species), data = iris) +
  ggtitle("Petal Data")
animint2gist(list(p1 = p1, p2 = p2))
```

tdhock:
Hey Kevin that looks like a good start thanks for the update!

kferris10:
Sorry, but I need a little bit more clarification with the multiple selection idea again.  [This is a decent example](http://bl.ocks.org/kferris10/raw/80e36178680b294b1c25/).

What do you think should happen when the user enters `6` then `4` for `cyl`?  The way I think about it, first the selector should update for the `6` level, then it would update for the `4` level.  In this case, I don't understand what allowing the user to enter multiple selection values does.  I would think it would be simpler to just have the user choose `6`, then choose `4` using a single selector.

tdhock:
Since the multiple selection variable cyl appears in the legend I assume
that first=list(4, 6, 8), right?

In that case the multiple selection widget should begin with 4 6 8 already
selected in the selection box.

After de-selecting one of those values (e.g. 6, after clicking on a geom
with clickSelects=cyl or clicking on the cyl color legend) the multiple
selection widget should show only the other values which are still selected
(in this example, 4 and 8).

kferris10:
So you're thinking that the widget should show which values are currently selected?  The way I was thinking, the user would just use the widget to update one level at a time.

tdhock:
For multiple selection variables the widget should definitely show the full
set of values which are currently selected.

kferris10:
I think this is almost ready to go.  [Example](http://bl.ocks.org/kferris10/raw/9148afcbf6ddde66a8ff/)

The two things I can think of that still need work

1. The widgets don't update when you update the selector through the plot or legend.  This should be fixable with the [updateOption(https://github.com/brianreavis/selectize.js/blob/master/docs/api.md#methods_options)] method in selectize, but I haven't been able to get it to work yet.

2. I'm pretty sure the single selectors should behave differently, but what do you guys think?

kferris10:
I believe that this takes care of the two issues I mentioned.  [Here's a working example](http://bl.ocks.org/kferris10/raw/48909d53aa2ed4a3aac6/).  

```
mtcars$cyl <- factor(mtcars$cyl)
mtcars$vs <- factor(mtcars$vs)
p1 <- ggplot() + 
  geom_point(aes(wt, mpg, colour = cyl, size = vs, 
                 clickSelects = cyl, clickSelects = vs), data = mtcars)
p2  <- ggplot() + 
  geom_point(aes(mpg, wt, colour = cyl, size = vs, 
                 showSelected1 = cyl, showSelected2 = vs), data = mtcars)

viz <- list(p1 = p1, p2 = p2, 
            selector.types = list(vs = "single", cyl = "multiple"), 
            first = list(vs = 0, cyl = c(4,6,8)))
animint2gist(viz, description = "Selector Widgets")
```

Also, @tdhock I think there might be a bug with the legend opacity.  The vs legend doesn't seem to have it until you toggle the values.

cpsievert:
Looking good @kferris10! Just FYI, [multiple clickSelects is implemented yet](https://github.com/tdhock/animint/pull/31), so that might be causing some of the problems that occur when clicking the left hand plot.

kferris10:
Thanks!  I've been looking at some of the test errors, and I have a few questions.

1. In `test-global-variables`, `jQuery111306578069811587629` is now showing up in `diff.vars`.  Is this because I'm including a reference to jQuery?  I'm not really sure what the point of this test is so I don't know what to expect here.

inst/htmljs/animint.js
```diff
@@ -1780,6 +1796,212 @@ var animint = function (to_select, json_file) {
...
+    var selector_table = element.append("table")
+      .style("display", "none")
+    ;
+    var selector_first_tr = selector_table.append("tr");
+    selector_first_tr
+      .append("th")
+      .text("Toggle selected value");
+      
+    //creating an array to contain the selectized items
+    jquery_thingy_array = [];
```
kferris10: 2. I omitted the var here to try and explain what is going on.  If I open the browser, and try to update the `vs` selector, I get this error message.  I'll keep poking around, but if anyone has an idea why this is happening, I would love to hear it.  Also, I apologize for the name.  

image

inst/htmljs/animint.js
```diff
@@ -1780,6 +1796,212 @@ var animint = function (to_select, json_file) {
...
+    var selector_table = element.append("table")
+      .style("display", "none")
+    ;
+    var selector_first_tr = selector_table.append("tr");
+    selector_first_tr
+      .append("th")
+      .text("Toggle selected value");
+      
+    //creating an array to contain the selectized items
+    var jquery_thingy_array = [];
```
kferris10: 2. If I do `var jquery_thingy_array = [];` then try to update the `vs` selector,  I get an error in line 1494 (where I modify `jquery_thingy_array` in `update_selector()`) saying that `jquery_thingy_array` is undefined.

3. If I delete the `var` and just do `jquery_thingy_array = [];` then try to update the `vs` selector, I get the error message below.  I'll keep poking around, but if anyone has an idea why this is happening, I would love to hear it.  Also, I apologize for the name `jquery_thingy_array `.  It will be updated as soon as I get this straightened out.

image

kferris10:
I believe that this takes care of almost everything.  You can disregard my previous comments, they mostly aren't relevant anymore.  [Working example](http://bl.ocks.org/kferris10/raw/df24bf32ce9937bc342e/)

```
mtcars$cyl <- factor(mtcars$cyl)
mtcars$vs <- factor(mtcars$vs)
p1 <- ggplot() + 
  geom_point(aes(wt, mpg, colour = cyl, size = vs), 
             data = mtcars)
p2  <- ggplot() + 
  geom_point(aes(mpg, wt, colour = cyl, size = vs), data = mtcars)
viz <- list(p1 = p1, p2 = p2, 
            selector.types = list(vs = "single", cyl = "multiple"), 
            first = list(vs = 0, cyl = c(4,8)))
```

I believe we are down to just two issues with the widgets!  (There are also a couple of bugs unrelated to the widgets that yuo may notice.  I'll take care of them as well).

1.  I still am a bit confused about global variables.  Is it okay that, in test-global-variables, jQuery111306578069811587629 is now showing up in diff.vars?  This is the version of jQuery that I used so is it showing up because of this?

2.  I'm still working on figuring out what to do when there's nothing selected (i.e. the user has deselected everything).  Hopefully that wont take too long.

Then I just have to update the `selector-widgets` test and we should be good to go.

cpsievert:
Looking great @kferris10! 

We'll have to address and fix (1), and if I were to guess, it may be that a variable is being defined without `var` (and thus is being introduced to the global environmnent).

kferris10:
This latest push takes care of the issue with everything de-selected.  Now we're just down to the global variables issue.

Here's the output I get from `diff.vars` in this test.

```
> tests_run(filter = "global")
global variables : Called from: eval(expr, envir, enclos)
Browse[1]> diff.vars
[1] "animint"                      "jquery_thingy_array"         
[3] "jQuery1113007219210239316454"
```

I have no idea why `jQuery1113007219210239316454` is in there.  My only guess is that it comes from jQuery itself?  If don't script jquery in in `index.html`, then it disappears from `diff.vars`.

The `jquery_thingy_array` is currently in there because I need to access it in both the `d3.json()` part of animint and in `update_selector()` and I can't figure out how to refer to it in both places without making it a global variable.  Does that make sense?

tests/testthat/test-selector-widgets.R
```diff
@@ -0,0 +1,56 @@
...
+  # testing that the vs widget is present
+  vs_widget <- getNodeSet(info$html, '//select[@id="vs_input"]')
+  expect_equal(length(vs_widget), 1)
+  # testing that the cyl widget is present
+  cyl_widget <- getNodeSet(info$html, '//select[@id="cyl_input"]')
+  expect_equal(length(cyl_widget), 1)
+})
+
+test_that("Updating cyl widget shows/hides points", {
+
```
kferris10: @cpsievert I could use your help here too.  Do you know how to interact with selectize through the test?

cpsievert: I think you could do this with RSelenium's `findElement` and `clickElement()` methods, but I've found that using `executeScript` is often faster. This page documents those and other handy methods -> http://www.rdocumentation.org/packages/RSelenium/functions/remoteDriver-class

I was able to execute this in my Firefox console for the example you link to above:

```js
// introduce the dropdown content of the 'vs' selector to the DOM
document.getElementsByClassName("selectize-input")[0].click();
// store the content
var con = document.getElementsByClassName("selectize-dropdown-content")[0]
// click on the value '1' (which happens to be the second node
con.childNodes[1].click()
```

kferris10: I couldn't figure how to click `con.childNodes` without it crashing.  This does run fine in my browser though.

```
1. Error: Updating cyl widget shows/hides points -----------------------------------
	 Summary: UnknownAn unknown server-side error occurred while processing the command.
 	 class: org.openqa.selenium.WebDriverException
1: withCallingHandlers(eval(code, new_test_environment), error = capture_calls)
2: eval(code, new_test_environment)
3: eval(expr, envir, enclos)
4: remDr$executeScript("\n    // introduce the dropdown content of the 'vs' selector to the DOM\n    document.getElementsByClassName('selectize-input')[0].click();\n    // store the content\n    var con = document.getElementsByClassName('selectize-dropdown-content')[0];\n    // click on the value '1' (which happens to be the second node)\n    con.childNodes[1].doubleclick();\n                      ") at test-selector-widgets.R:50
5: queryRD(paste0(serverURL, "/session/", sessionInfo$id, "/execute"), "POST", qdata = toJSON(list(script = script, 
       args = args)), json = TRUE)
6: checkStatus()
7: stop(errMessage, call. = FALSE)
```

tests/testthat/test-selector-widgets.R
```diff
@@ -0,0 +1,63 @@
...
+test_that("Updating cyl widget shows/hides points", { 
+  remDr$executeScript("
+    // introduce the dropdown content of the 'vs' selector to the DOM
+    document.getElementsByClassName('selectize-input')[0].click();
+    // store the content
+    var con = document.getElementsByClassName('selectize-dropdown-content')[0];
+    // click on the value '1' (which happens to be the second node)
+    con.childNodes[1].click();
+                      ")
+})
```
tdhock: can you please write an expectation for this test?

tdhock:
Hey @kferris10 this selector widget functionality looks like it is working exactly as I imagined it, on the WorldBank data viz that I looked at. Great job! Anything else to do before merging with master?

tests/testthat/test-selector-widgets.R
kferris10: I still need a little bit of help writing this test.  Maybe @cpsievert has some ideas why `con.childNodes[1].click();` returns an error when using `tests_run()`?

tdhock: instead of using `executeScript`, you may consider using `clickID`, `remDr$findElement`, `e$clickElement`, and `e$sendKeysToElement` as in test-widerect.R

cpsievert: @kferris10 

[Looks like `HTMLElementObject.click()` is (sort of) on the TODO list for phantomjs 2.0](https://github.com/ariya/phantomjs/issues/10795), but I couldn't get it to work locally (even worse, some other test crashes it)

I suppose we should go with @tdhock suggestion, but note that 
[those tests are skipped on phantomjs](https://github.com/tdhock/animint/blob/c95324784fecd04012afb101e2e3263c90c4be7f/tests/testthat/test-widerect.R#L231-L282), so you'll probably want to skip them here as well (and verify with firefox locally)

kferris10: Okay.  I'm pretty sure that I could replicate your example @tdhock in test-selector-widgets if that would be better.

tdhock: maybe your test in test-selector-widgets.R
<https://github.com/tdhock/animint/pull/96#discussion_r38374678> is no
longer necessary? I added the tests yesterday in test-widerect.R and that
seems to exercise most of the selectize features I would use.

kferris10: That sounds fine to me.  I can delete test-selector-widgets.R and merge it tonight.

tdhock: ok but we should take a look at fixing the other two test failures before merging.

kferris10: My bad.  I wasn't paying close enough attention.  I'll look at those as soon as I can

tdhock: I will have a look later today too.

caijun:
The shiny test should be disabled due to the problem of PhantomJS. https://github.com/ariya/phantomjs/issues/11526

cpsievert:
That's a bummer, go ahead and re-disable.

kferris10:
This takes care of the malaria tests for me.  It was the same problem we had with some of the malaria tests where I forgot to remove periods from variable names https://github.com/tdhock/animint/commit/2349c9b6067dee34d0fb03df4bb313a1be73df3f

Sorry this took a few days.  If @cpsievert could tell the Twins to stop winning, that might help me find some more free time :)

tdhock:
Thanks for the fixes @kferris10 

Tests pass on my machine using firefox.

I updated the version and will merge with master.

Issue #89, opened 2015-06-22
Some themes from the ggthemes package don't seem to work


kferris10:
The [ggthemes](https://github.com/jrnold/ggthemes) package provides some additional themes for ggplot2.  But some of them don't seem to work with animint.

```
animint2dir(list(p1 = qplot(Sepal.Width, Sepal.Length, data = iris, colour = Species) + theme_fivethirtyeight()))

 Error in serialize(object, connection = NULL, ascii = ascii) : 
  object 'direction' not found 
```

Traceback shows that the error occurs when using `guides_train()` in `getLegendList()`

```
4 ggplot2:::guides_train(scales = scales, theme = theme, guides = guides.list, 
    labels = plot$labels) at animint.R#1127
3 getLegendList(meta$built) at animint.R#209
2 parsePlot(meta) at animint.R#929
1 animint2dir(list(p1 = qplot(Sepal.Width, Sepal.Length, data = iris, 
    colour = Species) + theme_fivethirtyeight())) 
```

This error occurs for the following themes:

* `theme_calc()`
* `theme_fivethirtyeight()`
* `theme_gdocs()`
* `theme_wsj()`

kferris10:
Closed with #95

PR #95, opened 2015-07-03
Fix grid lines

kferris10:
Fix bug with the grid lines when no horizontal-minor grid lines where present.  [The updated tutorial](http://kferris10.github.io/animint/#scale_y_log10-and-geom_contour-with-stat_density2d) looks pretty good to me.

[There's still a problem](http://kferris10.github.io/animint/#geom_abline) with multiple ablines that I will fix momentarily.

PR #97, opened 2015-07-09
Fix bug where theme$legend.direction accidentally becomes NULL

kferris10:
I think there's a bug with the way that `is.null()` is being used sometimes.  The following example is typical of some of the code in `animint.R`.  I think the goal is for `x` to stay as 3, but that's not what R is doing.

```
x <- 3
x <- if(is.null(x)) 6
x
NULL
```

I think that it might make more sense for the code to be written this way.  

```
x <- 3
if(is.null(x)) x <- 6
x
[1] 3
```

Doing this will also help with #89

### DIFF
--- a/R/animint.R
+++ b/R/animint.R
@@ -1096,9 +1096,10 @@ getLegendList <- function(plistextra){
   theme$legend.key.width <- if(is.null(theme$legend.key.width)) theme$legend.key.size
   theme$legend.key.height <- if(is.null(theme$legend.key.height)) theme$legend.key.size
   # by default, direction of each guide depends on the position of the guide.
-  theme$legend.direction <- if(is.null(theme$legend.direction)){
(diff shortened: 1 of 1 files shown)


kferris10:
It looks like the Travis build had a problem with `animint2gist`.  I didn't have any issues when testing on my computer so if anybody could make a suggestion towards what is wrong, that would be great.

cpsievert:
Oops, I generated a new personal access token for #95 and deleted the old one (that is why the gist test doesn't work). I suggest we focus on merging #95 ASAP, then returning to this.

IIRC, this style of code (conditionally overwriting a certain variable) is in a lot of different places in the compiler. If @tdhock is OK with it, I'd prefer if we use this style whenever possible.

```r
"%||%"  <- function(x, y) {
  if (length(x) > 0) x else y
}
x <- NULL %||% 6
```

cpsievert:
I merged these commits into #95

PR #98, opened 2015-07-14
Fix bug with Geom abline

### LINKED ISSUE #92: geom_abline doesn't work

kferris10:
Pretty sure that this should take care of all the problems with `geom_abline()` and close #92 

[See here for fixed example](http://bl.ocks.org/kferris10/raw/8cb4ee8562e204eac381/)

### DIFF
--- a/R/animint.R
+++ b/R/animint.R
@@ -94,7 +94,9 @@ parsePlot <- function(meta){
       ##    theme.pars$panel.grid then from theme.pars$line
       for(i in 1:length(pars)) {
         if(is.null(pars[[i]])) pars[[i]] <- 
-          if(!is.null(theme.pars$panel.grid[[i]])) {
+          if("element_blank" %in% class(theme.pars$panel.grid)) {
+            theme.pars$line[[i]]
+          } else if( !is.null(theme.pars$panel.grid[[i]]) ) {
             theme.pars$panel.grid[[i]]
           } else {
             theme.pars$line[[i]]
--- a/R/animint.R
+++ b/R/animint.R
@@ -92,11 +92,9 @@ parsePlot <- function(meta){
       
       ## if elements are not specified, they inherit from 
       ##    theme.pars$panel.grid then from theme.pars$line
-      for(i in 1:length(pars)) {
+      for(i in names(pars)) {
         if(is.null(pars[[i]])) pars[[i]] <- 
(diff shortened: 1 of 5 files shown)


cpsievert:
Looks good to me! +1 
cc @tdhock

R/animint.R
```diff
@@ -94,7 +94,9 @@ parsePlot <- function(meta){
       ##    theme.pars$panel.grid then from theme.pars$line
       for(i in 1:length(pars)) {
         if(is.null(pars[[i]])) pars[[i]] <- 
-          if(!is.null(theme.pars$panel.grid[[i]])) {
+          if("element_blank" %in% class(theme.pars$panel.grid)) {
+            theme.pars$line[[i]]
+          } else if( !is.null(theme.pars$panel.grid[[i]]) ) {
```
kferris10: I just tacked this change on to here because it's pretty small.  Let me know if this still looks okay.

tdhock: can you please add a comment that explains what you are trying to do? ideally add a reference to the appropriate test.

kferris10: What I want to do is say "if the parameter is NULL then look for it in `theme.pars$panel.grid`, if it's NULL there then look for it in `theme.pars$NULL`.

There's an error, however, if `$panel.grid` is an empty list - i.e. if it's created by `element_blank()`.
```
x <- list()
x[[1]]
Error in x[[1]] : subscript out of bounds
```

So I believe I first need to check whether it's specified as `element_blank()` or not.  But is there a better way to do this?

tdhock: there is the `is.blank` helper function which maybe would help clarify the code https://github.com/tdhock/animint/blob/master/R/animint.R#L142

kferris10: I think this is a better solution than my previous attempt.  While this generates an error
```
x <- list()
x[[1]]
Error in x[[1]] : subscript out of bounds
```
this returns `NULL` like I want
```
x[["colour"]]
NULL
```

cpsievert:
Looks OK to me +1

tdhock:
+1

PR #102, opened 2015-08-03
Fix interactive legend bugs

tdhock:
Hey @kferris10 I was using your new interactive legends for a data viz collaboration with a colleague @zmartine, and we noticed that there are a couple of bugs which I now check for in test-malaria.R (named after the data set we are working on).

Can you please have a look at fixing these bugs when you get a chance?

- [x] yellow geom_segment (region.type=LCR) does not render initially. (but does render after clicking that legend entry)
- [x] the legend entry text should dim to opacity=0.5 after clicking to de-select a value. This works for the annotation legend but not any of the other legends. Any idea why?
- [x] When we specify `geom_point(aes(fill=annotation, color=highly.divergent.regions))` the color legend should show a different stroke but the same fill for each point. This is not technically a bug with your interactive legends code, so I will look into fixing it myself via changing the compiler R code that deals with legends.
- [x] At least in firefox I don't see any visual clue that the legend entry items are clickable. Upon hovering over each legend entry would it be possible to add a grey border or fill of the `<tr>`? And then remove the grey (set to white) when the mouse is moved away from the `<tr>`?

By the way, the interactive legends are really nice for this particular data viz, thanks again for implementing that! In particular it was nice that I could easily add a bunch of showSelected aesthetics to the third plot so that objects appear/disappear after clicking the legend in the second plot.


tdhock:
here is a link to the rendered data viz http://bl.ocks.org/tdhock/raw/17cdd84fe2dbe19ac17f/

kferris10:
Hey Toby, I'm glad that you're finding it useful. I just moved and don't have access to internet yet, so I won't be able to take a look at this problem until the weekend. But I'll take a look as soon as I can.

Just an update, the widgets are going well. I think they're almost finished (again, can't post anything without Internet, but that should be fixed soon)!

tdhock:
here is the updated data viz with the white fill in the color legend points http://bl.ocks.org/tdhock/raw/2e27a66d4f83217c9add/

kferris10:
I believe the failure with legend is coming because your variables have `.` in the name.  I'm guessing, but when we try to select the legend entries to update opacity, the selector is getting confused and mistaking the dots for classes.  I only had time to check `highly.divergent.regions` right now, but just changing the name to `highly_divergent_regions` fixed the problem.
https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L1493

To fix this, I think that I could just change the variable name to have underscores when creating the legend class.  Do you see any issues with this?

kferris10:
@tdhock When I view examples in Firefox, I get the pointer suggesting that they are clickable.  Could you double check just to make sure?

tdhock:
Hey @kferris10 thanks for contributing these fixes. The interactive legend functionality works great! The relevant tests pass on firefox on my machine, and I just pushed a commit which should fix the shiny tests on wercker. I will update NEWS/DESCRIPTION and then merge with master.

Issue #103, opened 2015-08-07
Legend interactivity not working in R markdown


cpsievert:
Legend interactivity [works fine](http://bl.ocks.org/cpsievert/raw/f8981104d61021374efd/) when creating a standalone page:

```r
library(animint)
p <- qplot(data = iris, x = Sepal.Width, y = Sepal.Length, color = Species)
animint2gist(list(p = p))
```

However, something goes haywire when I try to [embed in R markdown](https://gist.githubusercontent.com/cpsievert/732835a05da8b741b3a3/raw/a8c6ebddd81bd021c3ff32a662367481ebeb71eb/index.Rmd). 

Any thoughts, @kferris10 ?

kferris10:
I actually encountered this issue too when trying to put together a blog post about legend interactivity.  Just to check, when you compile it you get a folder in your directory for that chunk, right?  And when you do `servr::httd("unnamedchunk1")` it works fine right?

cpsievert:
yup

kferris10:
I'm not sure how rmarkdown interacts with animint.  Could you point me to where in the code that takes place?

cpsievert:
When you print an object in knitr (`rmarkdown::render()` eventually calls `knitr::knit()`) it calls `knitr::knit_print()`on that object. [Here](https://github.com/tdhock/animint/blob/f7d47b5a9b019fc3b12c0aa2d822572faf7ce4df/R/knitr.R#L15-L44) we tell knitr how to print objects with a class of animint

kferris10:
Moved to #107

PR #107, opened 2015-09-06
Legend interactivity with knitr

kferris10:
@cpsievert I was able to get this working on my machine ([working example](http://kferris10.github.io/test-animint/) ).  I just needed to fix the selection of legend rows.

I can't seem to get the test for this working though.


cpsievert:
When I click on a legend entry, in my browser console, I'm getting:

```
Uncaught TypeError: Cannot read property 'type' of undefined
```

Which references [this line](https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L1473)

kferris10:
The minimal example that you posted in #103 works fine though, right?

I'm pretty sure that the problem is that it's not using the right `Selectors` object.  For some reason, it's using the one for the second plot (which is empty).

cpsievert:
Yep, looks as though it works when there is a single animint plot, but not when there are multiple...

kferris10:
Just checking, do you know if interactivity for multiple plots worked before this summer?  I'm assuming this is something that I changed (might have to do with the widgets), but that would be nice to know.

cpsievert:
I remember legend interactivity working for multiple plots within a bare `animint2dir()` page, but I don't think we ever got multiple _independent_ plots within an Rmd document working

kferris10:
Okay.  Do we want to try and fix that before merging this branch?  I think we've figured out the knitr problem

tdhock:
I seem to recall Carson's worldPop Rmd viz working with several animints on the same Rmd-generated html page, but now the link is broken http://cpsievert.github.io/animint/worldPop/worldPop.html

I would recommend fixing that before merging this branch.

cpsievert:
I don't think that missing link has to do with anything we've done since the summer. I'm pretty sure it's just a problem with my website.

tdhock:
sorry for the confusing comment.

i meant to say it would be nice to get the "multiple animints in a single Rmd" feature working again.

but it would also be nice to have that web page back up somewhere. do you have a copy of it somewhere?

cpsievert:
yea, the weird thing is the files haven't changed for over a year, but for some reason they aren't being served by GitHub right now -> https://github.com/cpsievert/cpsievert.github.com/tree/master/animint/worldPop

inst/htmljs/animint.js
```diff
@@ -1648,7 +1648,7 @@ var animint = function (to_select, json_file) {
     }
     
     // selecting points based on legend
-    d3.select("#plot").selectAll("#legend").selectAll("tr")
+    d3.selectAll("#legend").selectAll("tr")
```
tdhock: I think the problem with rendering multiple animints on the same web page may be related to the fact that you are currently using `d3.selectAll` here. I think it would make more sense to use `element.selectAll` or `svg.selectAll` (depending on if you want to select all legend entries in the entire animint or plot). Remember we have these definitions:

```javascript
  var element = d3.select(to_select);

    var svg = tdLeft.append("svg")
      .attr("id", p_name)
      .attr("height", p_info.options.height)
      .attr("width", p_info.options.width);
```

tdhock:
I added another animint to the Rmd test. Now there are two plots with interactive legends for a "col" variable. I discovered another bug: the first selectize widget works fine, but the second renders as a `<select>` element. @kferris10 any idea why?

screenshot-rmd-selectize-bug

kferris10:
Yes, I assigned each selectize widget an id of `variablename_input`.  Since the variable names are the same, I assume that selectize is just getting confused.

The id for the first plot

image


The id for the second plot

image

So I believe all we need to do is change [this line](https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L2065) and [this one](https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L2107) to `var $temp = $('.' + plotid + ' #' + s_name_id + "_input")`.  How can I tell what the plotid in the div will be though?

kferris10:
Sweet!  Glad this is working.

PR #109, opened 2015-09-16
bugfix for plots with many (>40) facet rows or columns

tdhock:
screenshot-many-facets-bug

Scales behave improperly when there are many facets. For example when there are many (>40) rows, any x values get scaled to negative values, so only the axes and strips (but no data) are rendered.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: animint
 Maintainer: Toby Dylan Hocking <tdhock5@gmail.com>
-Author: Toby Dylan Hocking, Susan VanderPlas, Carson Sievert
-Version: 2015.09.08
+Author: Toby Dylan Hocking, Susan VanderPlas, Carson Sievert, Kevin Ferris, Tony Tsai
+Version: 2015.09.16
 License: GPL-3
 Title: Interactive animations
 Description: An interactive animation can be defined using a list of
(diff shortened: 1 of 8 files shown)


tdhock:
The problem can be seen in the screenshot below for which firebug stopped when rendering the first plot. `graph_height` is negative since it is computed by subtracting `strip_heights.reduce(...)` which is very big since `strip_heights` is an array of length 40: `[14, 14, 14, ...]`. @cpsievert is it correct that the length of `strip_heights` should be 40, even in this first plot which has 1 row and 40 columns?

screenshot-many-facets-firebug

Same thing for the second plot: `graph_width` is negative since it is computed by subtracting the sum of `strip_widths` (another array of size 40).

tdhock:
Also I am seeing this test failure on phantom/travis, but not on firefox on my computer:

```
1. Failure (at test-many-facets.R#64): right strips all at the same x position -
length(text.list) not equal to n.circles
40 - 0 == 40
```

For some reason `//svg[@id="vertical"]//g[@class="strip"]//text` works fine for selecting 40 `<text>` elements on my computer but not on travis. Do you have any idea why? If you have a minute, could you please check if you see if you get the same error on your computer with phantomjs or firefox?

cpsievert:
Are you using `facet_grid()` in that example? I'm pretty sure there is room for strip labels on each panel (which makes sense for `facet_wrap()`, but not `facet_grid()`). For a proper fix, we'll probably have to send the type of facet to the renderer and conditionally make room. I'm leaving for Australia tomorrow, and I have a busy schedule through September, so I might not have enough time to fix this for a while :/

tdhock:
thanks for the advice, no problem if you are too busy to check it out right now. I should be able to figure it out eventually.

PR #115, opened 2015-10-07
do not render widgets for .variable .value selectors

tdhock:
There can be arbitrarily many selectors, and typically the only ones that I want to use are those not defined by .variable/.value aesthetics. After profiling this animint http://bl.ocks.org/tdhock/raw/c15cd081f6856641cda5/ in chrome I found that the slow loading was caused by rendering of selectize widgets with many values. So I propose to decrease the animint loading time in animints like this, we do not render the selectize widgets for .variable/.value selectors.

I implemented a test that is failing in test-variable-value.R, and the fix should involve slight modifications to the compiler and renderer.


kferris10:
I'm a little bit confused.  Do you mean that if the showSelected variable is named `something.value` or `thing1.variable` then the widgets should ignore it?

tdhock:
hey @kferris10 sorry for the confusing comment. Please refer to https://github.com/tdhock/animint/wiki/Advanced-features-present-animint-but-not-in-ggplot2#data-based-selector-names-using-variable-value-aesthetics for an explanation of the .variable .value aesthetics feature. So actually it is for `aes(showSelected.variable=something.var, showSelected.value=something.val)` which will generate as many selectors as there are unique values of `something.var`.

tdhock:
TODO make https://github.com/tdhock/malaria-evolution-viz/blob/master/figure-folds.R into a test (click to change test.fold), fix JS errors.

kferris10:
I think I understand now.  Do you need any help with this?  I just briefly glanced through the changes so I'm not really sure what still needs to be done.

tdhock:
All tests pass on my machine using firefox.

tdhock:
TODO:
- [ ] test for aes(hjust, vjust)
- [ ] error for showSelected=variable with no time or clickSelects.
- [ ] what happens when showSelected=var showSelected2=var?

tdhock:
all tests pass on my machine using firefox.

R/animint.R
```diff
@@ -1285,6 +1350,46 @@ animint2dir <- function(plot.list, out.dir = tempfile(),
...
+    has.legend <- sapply(meta$selectors, function(s.info)isTRUE(s.info$legend))
+    is.trivial <- one.level & (!has.legend)
+    if(any(is.trivial)){
+      ## With the current compiler that has already saved the tsv files
+      ## by now, we can't really make this data viz more efficient by
+      ## ignoring this trivial selector. However we can warn the user so
+      ## that they can remove this inefficient showSelected.
+      warning("showSelected variables with only 1 level: ",
+              paste(names(meta$selectors)[is.trivial], collapse=", "))
+    }
```
tdhock: Hey @kferris10 thanks for the idea about the possibility of a showSelected variable with only 1 level. In that case we warn the user if it did not come from a legend with 1 entry (which is also a pretty weird situation, but possible, as in `aes(fill=place)` in https://github.com/tdhock/animint/blob/master/tests/testthat/test-renderer2-interactivity.R)

tdhock:
So @kferris10 I propose to merge this PR now and then if you want to clean up my use of three underscores, can you please submit a separate PR?

kferris10:
That sounds fine.  I actually looked through the underscores yesterday evening and couldn't figure out a better approach (not saying one doesn't exist, but I couldn't figure it out).  Do you have any problem with just leaving it the way it is?

tdhock:
no I don't have any problem with the three underscores implementation. The only problem that I was having was slow page loading when there were many (>10) selectize widgets to render, but this PR fixes that.

PR #116, opened 2015-10-23
PredictedPeaks data and tests

tdhock:
I was designing the data viz below and I noticed that animint.js is sometimes rendering twice as many `<text>` elements as necessary in the first/top plot. In the screenshot below we see 3 text elements, and there are only 3 rows specified in the data:

```r
> subset(PredictedPeaks$chromCounts, chrom=="chrM" & dotID=="38 neutro samples, 1 Input samples")
       nonInputType   type chrom chromStart chromEnd        peak.name
259579       neutro  Input  chrM      16039    16155 chrM:16039-16155
259580       neutro   mono  chrM      16039    16155 chrM:16039-16155
259581       neutro neutro  chrM      16039    16155 chrM:16039-16155
       samples.up type.fac up Input                              dotID bases
259579          1    Input 38     1 38 neutro samples, 1 Input samples   116
259580         14     mono 38     1 38 neutro samples, 1 Input samples   116
259581         38   neutro 38     1 38 neutro samples, 1 Input samples   116
       middle zoom.bases zoomStart zoomEnd relative.middle
259579  16097       1160     14937   17257       0.9715131
259580  16097       1160     14937   17257       0.9715131
259581  16097       1160     14937   17257       0.9715131
> 
```

screenshot-predictedpeaks

but in fact when I inspected the HTML in FireBug there are 6 elements!

I managed to figure out that `update_selector` is being called twice after clicking a chromosome in the second plot, and the second call ends up creating twice as many elements. I commented out these lines of the JavaScript code, and the problem went away:

```javascript
      // update selected widgets, if necessary
      if(s_info.type == "multiple" | 
	 selectized_array[v_name].getValue() != selected_ids) {
	selectized_array[v_name].setValue(selected_ids);
      }
```

So that tells me that `update_selector` is being called again inside of the selectize widget's `getValue` or `setValue`. @kferris10 do you have any ideas about how to fix this?


tdhock:
Also there is a second data viz in test-renderer2-PredictedPeaks.R but we can not render the selectize widget for the `peak.name` and dotID variables, too slow. Any idea why? (I edited the animint.js code so that it does not render a widget if there are more than 1000 levels, but can you see if you can remove that bit of code and see if you can speed up the rendering?)

kferris10:
Hey @tdhock, sorry about this.  I thought I outlined the logic behind calling update_selector twice somewhere, but I forget where.  

The issue is that I need to call update_selector from within the widget.  But if a selector is updated from somewhere other than a widget than I need to call the widgets to update the selected values.  So if you update a widget, it calls update_selector which then tries to setValues for the widget which activates the widget.  It's redundant, but I wasn't able to think of a better way to do it.  Does that explanation make sense?

I'll take a more thorough look at your changes this afternoon.

tdhock:
Hey @kferris10 thanks for the input. No problem, just wondering if you knew about a quick fix. After googling about the selectize setValue method, I found that by setting its second argument, we can avoid calling its onChange method, and thus avoid calling update_selector twice. So that bug is fixed.

However the second issue is still unsolved: slow rendering of a selectize widget with many values. For now I propose that we decide at compile time whether or not to render the selectize widgets. By default if there is a selector with more than 1000 values then the compiler does not set the `levels` and so the renderer will not create a selectize widget. However this can be overridden by a new `selectize` option.

tdhock:
docs for the new selectize option https://github.com/tdhock/animint/wiki/Advanced-features-present-animint-but-not-in-ggplot2#use-selectize-option-to-showhide-selector-widgets

inst/htmljs/animint.js
```diff
@@ -1692,11 +1691,11 @@ var animint = function (to_select, json_file) {
...
-      // update selected widgets, if necessary
-      if(s_info.type == "multiple" | 
-	 selectized_array[v_name].getValue() != selected_ids) {
-	selectized_array[v_name].setValue(selected_ids);
-      }
+      // from
+      // https://github.com/brianreavis/selectize.js/blob/master/docs/api.md:
+      // setValue(value, silent) If "silent" is truthy, no change
+      // event will be fired on the original input.
+      selectized_array[v_name].setValue(selected_ids, true);
```
tdhock: Seems to me like the `if(multiple selection or selectized.getValue is not selected_ids)` is not necessary here. @kferris10 any reason you can think of for keeping this?

tdhock:
I moved the PredictedPeaks test data set to https://github.com/tdhock/animint-examples/tree/master/data since it is over 5MB even with `compress="xz"`

I am ready to merge with master, but I will wait for a +1 from @kferris10. Anything you think I should change or clarify before merging?

tdhock:
TODO add a simple test for `aes(color="constant")` since this seems to cause a problem with the current implementation of interactive legends.

inst/htmljs/animint.js
kferris10: I think that looks great.  Makes way more sense actually.

kferris10:
It all looks good to me.  Anything that I can help with?

tdhock:
@kferris10 there is one more thing I wanted to add to this PR: stop with an error at compile time if `aes(color)` is not a variable name (and also do the same for other aes that would generate an interactive legend). Two cases are `aes(color="constant")` and `aes(color=factor(numeric.variable))`. Currently `aes(color="constant")` does not work, and `aes(color=factor(numeric.variable))` makes a selection variable called `show_color`. I think it would be better to stop with an error and have the designer manually create an informative selection variable name other than `show_color`. This would also simplify the compiler code. Would that be OK with you?

tdhock:
all tests pass on firefox on my computer

kferris10:
Why do you want to stop with an error?  If you don't like the `show_colour` name, we could just change it to `automatic_show_color`.  That might be a good idea anyway.  Or we could generate a warning message.  But I don't know if we need to error out.

tdhock:
Thanks for sharing your thoughts Kevin. Here is my line of thinking. Recently I noticed that interactive legends in one plot can be really useful when combined with `aes(showSelected)` in another plot. For example in https://github.com/tdhock/animint/blob/PredictedPeaks/tests/testthat/test-renderer2-VariantModels.R I used `aes(color=method)` in the second plot with `aes(showSelected=method)` in the first plot. It would have been problematic to use `aes(color=factor(method))` in this case since it would have generated a legend with interactivity for the `show_color` selection variable. But we wanted to use the `method` selection variable. To avoid confusion in cases like this I think it is better to just stop the compiler with an error, and instead require the designer to explicitly specify what selection variable they intend to be linked to that legend.

Another example (that I just thought of and probably should be added as a test) is when you have one plot with `aes(fill=some.variable)` in one plot and `aes(fill=a.completely.different.variable)` in another plot. It would not make sense to link both fill legends with a common `show_fill` variable, right?

tdhock:
you are right that we *could* implement a method for analyzing expressions and computing a selector variable name automatically. An easy case would be `aes(color=factor(var1))` (just use `var1` as the selector) but the harder cases would be like `aes(color=ifelse(var2 > 0, "positive", "negative"))` -- should we infer a selector name of `var2` in that case? I do use expressions like this quite frequently, too. But they can always just be avoided by adding another column to the data.frame.

Also, from my perspective as the lead developer of animint, it is much easier to maintain a compiler that just stops with an error (rather than implementing some complicated expression analysis algorithm).

kferris10:
Sounds good to me.

PR #119, opened 2015-11-03
compiler should stop with an error on multiple showSelected

tdhock:
when we have `aes(showSelected=var1, showSelected=var2)` animint should stop with a clear error message which indicates that is a problem. Currently we have

```
Error in `[.data.frame`(g.data, , can.chunk.cols, drop = FALSE) : 
  undefined columns selected
```


tdhock:
also it seems like we need a test for animint with multiple chunk vars 
screenshot-animation-exception

tdhock:
also the compilation is painfully slow for that data viz, maybe we could speed up the common chunk code?

tdhock:
I saved the data set here https://github.com/tdhock/animint-examples/blob/master/data/ChromHMMinit.RData?raw=true and the code https://github.com/tdhock/ChromHMM-viz/blob/master/figure-repeats.R can be adapted for a test

PR #135, opened 2016-03-04
depend ggplot2 >=2.0

tdhock:
hopefully we can get animint to work with the new ggplot2 internals one day...


kferris10:
A visualization of the current ggplot2 internals from @cpsievert [link here](https://cpsievert.shinyapps.io/ggtree/)

image

And here is the old ggplot2 internals [link](https://kferris.shinyapps.io/ggtree/)

image

kferris10:
A few things that I notice

- `ggplot$plot` is mostly the same.  The only difference I see is `plot$ordinates`
- `ggplot$data` is now an object of lenth two.  I'm guessing this means there is now a data object for each geom layer.  Might be one for each stat too.
- `ggplot$panel$ranges`, `$panel$shrink`, and `$panel$layout` look like the are the same as before.  It looks like `$panel$x_scales` and `$panel$y_scales` now inherit from somewhere else.

I'm guessing that this means that we'll have to rewrite animint so that it pulls data from the appropriate layer and so that it sets x and y scales correctly.  If I've missed anything, please feel free to chime in.

tdhock:
Thanks for your analysis Kevin. Looks like it should not be so hard after all if there are so many similarities. I guess the next step is to look at where the compiler is yielding errors, put in some `browser()` calls, and figure out what is going wrong e.g. https://app.wercker.com/#buildstep/56d9bf501618a4fe2c098b06

tdhock:
some info from Carson, quoting:

I've been putting a lot of time into plotly::ggplotly() as you might have saw here -- https://github.com/ropensci/plotly/pull/480

There is a lot in that pull request which goes well beyond fixes for changes in ggplot2 internals. It's true that the data structure returned by ggplot2::ggplot_build() hasn't changed much, but the internals which create that output have changed a lot. Most importantly, Winston created a new OO system to replace proto, named ggproto. I highly recommend reading this to get acquainted with it -- https://github.com/hadley/ggplot2/blob/master/vignettes/extending-ggplot2.Rmd

I put a lot of time into proper sizing/styling for theme elements -- animint could probably benefit from this unitConvert function that I use for converting arbitrary grid units to pixels or a 0-1 coordinate system -- https://github.com/ropensci/plotly/blob/bde1158/R/ggplotly.R#L534-L564

There is also initial support for accessing plotly events from shiny -- here are some examples:

http://104.131.111.111:3838/plotlyEvents/
http://104.131.111.111:3838/plotlyLinkedClick/
http://104.131.111.111:3838/plotlyLinkedBrush/
https://twitter.com/kyle_e_walker/status/706961730459582464

kferris10:
@tdhock finally got a chance to start looking at the new ggplot2 internals.  @cpsievert is right about having to fix more than just the `ggplot_build()` internals as some of the objects have changed so I no longer know how to work with them.  I just made a couple hacky commits so that we'd have something to work with.  [As vanilla as possible example](http://bl.ocks.org/kferris10/40e39db4c67895d83d3e) of an animint graphic that works, but doesn't do anything.

```
library(ggplot2)
library(animint)

p1 <- qplot(Sepal.Length, Sepal.Width, data = iris, colour = Species)

animint2dir(list(p = p1), 
            out.dir = "~/kevin-projects/test-animint")
```

I saw your other comment about moving forward with the current syntax so I'll keep plugging away with this (unless you've changed your mind)

R/animint.R
```diff
@@ -349,8 +349,11 @@ hjust2anchor <- function(hjust){
 #' @return list representing a layer, with corresponding aesthetics, ranges, and groups.
 #' @export
 saveLayer <- function(l, d, meta){
+  get_obj_gg2_class <- function(obj) {
+    tolower(substr(class(obj)[[1]], 5, nchar(class(obj)[[1]])))
+  }
```
cpsievert: You could steal my approach in plotly

```r
> plotly:::ggtype
function (x, y = "geom") 
{
    sub(y, "", tolower(class(x[[y]])[1]))
}
<environment: namespace:plotly>
```

tdhock:
Great to see that you got a plot to render, even if the interactive legend doesn't work.

About the syntax I think it is best to keep it the same for now. Ideally, we should be able to get all of the existing test cases to pass using the new compiler/renderer code.

kferris10:
@tdhock very slow progress, but these two example both work (though they look slightly different than expected).

```
p1 <- qplot(Sepal.Length, Sepal.Width, data = iris, colour = Species))
p2 <- ggplot(iris, aes(Sepal.Length, Sepal.Width)) + 
  geom_point(aes(clickSelects = Species))
```

tdhock:
Great thanks for the info Kevin. @faizan-khan-iit has been accepted for this year's GSOC so welcome to the Animint dev team! Kevin has been working in this branch on getting animint working with the new ggplot2 internals. Faizan, please take a look at Kevin's changes and ask him if you have any questions. Also I will add you as a collaborator so feel free to push any changes to this branch if they help get more tests to pass.

kferris10:
@faizan-khan-iit that's great to hear.  I've made a little progress, but it's been slow work.  Please let me know if you have any questions and I'll respond as quickly as possible.  

One quick note, the changes in commit [f0a2c4f](https://github.com/tdhock/animint/pull/135/commits/f0a2c4fab556b7ee4aa765edb24ce18dbcaa615e) were super hacky and will need to be updated at some point.  I just wanted to put something simple, that worked, together.

faizan-khan-iit:
@kferris10 @tdhock Thanks! I will get to it as soon as i possibly can.

@kferris10: I have a basic question. As we will be working with the new ggplot2 version, I need to install the newer version along with the older version(v1.0.1) on my Windows machine. I found [this](http://stackoverflow.com/questions/2988559/how-do-you-use-multiple-versions-of-the-same-r-package) as a possible solution.

Should I proceed with this or is there another way to set it up?

kferris10:
@faizan-khan-iit that looks reasonable.  If that becomes complicated, you could always just install whatever version of ggplot2 you will be working with when you start working (that's what I'm doing).

faizan-khan-iit:
@kferris10 I went through the discussion above. From what I understood, it seems a considerable part of the compiler needs to be edited. We could discuss the details of the task and make a road-map to follow.

But before that I needed some help. I went through your commits and wanted to test them on my machine. Could you give me a quick rundown of how we will be testing the compiler as we edit it? Also will I have to install _animint_ from this branch to bypass the _ggplot2_ dependency issues?

Thanks for your help!

kferris10:
@faizan-khan-iit sorry for the late reply.  I'll try to be more on top of things going forward, but feel free to follow up if I'm being slow.

For testing, @cpsievert will be a much better reference but I can try answering your questions.  What OS are you using?  I'm on Windows, and I could never get phantomjs to run.  After a lot of trial and error, I came up with this solution:

```
library(RSelenium)
library(testthat)
library(animint)

tests_init("firefox")
tests_run(filter = "plot-names")
tests_end()
```

Currently, I just install the appropriate version of ggplot2, then install whichever version of animint I want to work on.  But I'd just go with whatever workflow works for you.

faizan-khan-iit:
@kferris10 Thanks for the info!

I have R set up on both Windows and Ubuntu so I could work on any one.

Currently I have this setup that I thought would be best for Windows - I load the two versions (v1.0.1 and v2.1.0) of `ggplot` in two different workspaces and source the `animint` code and run the example through the debugger to see the workflow of the code. I am still getting familiar with the code so this helps a lot.

Also, the `qplot` example now works, after I found out that it was just an indexing error.

faizan-khan-iit:
Also I need some help [here](https://github.com/tdhock/animint/pull/135/files#diff-fddb8085c2de24f7941a89d932954d47L699).

I was looking at the edits you made here to rescale the axes. As you mentioned, this need to be modified for cases when the data frame does not have both `x` and `y` columns, or has some other variable that needs to be rescaled.

But I could not figure out how the old code worked as I can't seem to find any documentation for the `coord_transform` function. Does it work like the `coord_trans` function?

It would be helpful if you explained this bit of code here.

faizan-khan-iit:
I made some changes based upon my understanding, and I have made some progress. 

- [Kevin's Sepal example](http://bl.ocks.org/faizan-khan-iit/9eb0133b5bc2b447331a1a13ecb37639) looks better now. No work on legends yet.
- Also the [Lilac Chaser example](http://bl.ocks.org/faizan-khan-iit/035cb159f23f3469abef710ec2ce80ae) renders exactly like [before](http://bl.ocks.org/faizan-khan-iit/2a6fa9c6c93c8fd752b9).
- The [Grid Visual Illusion example](http://bl.ocks.org/faizan-khan-iit/8bb44a91c714f56a3100646046596adf) is almost identical to the one produced [using previous ggplot version too](http://bl.ocks.org/faizan-khan-iit/fa3a3d71be343267c27a).

That being said I would like someone to go through the commits once. I made [this commit to rescale data in data frame](https://github.com/tdhock/animint/pull/135/commits/0428880e9eea2f2588fa74b59fcbac0a5eb6673c) to pass more cases, but this does not seem like a solution. I would like some advice on this one.

The newer version of ggplot uses points (`pt`) instead of `lines`. Is there a conversion factor for this? I just used the one I deemed suitable, after looking at some examples.

kferris10:
@faizan-khan-iit Unfortunately I'm not sure how helpful I can be.  I wasn't able to figure out how the `coord_transform` function is supposed to work in the new ggplot2.  It is different from `coord_trans`.  @cpsievert do you have any good examples of how you work with `coord_transform` in plotly?

cpsievert:
I currently just take x/y positions from the built object (`panel$ranges`). If, and when, we implement polar coordinates, I'll probably need to use something like `CoordPolar$transform` to transform positions to polar coordinates, but I'm not sure why (or even if) we need to transform coordinate systems in animint.

R/animint.R
```diff
@@ -1684,6 +1692,28 @@ getLegendList <- function(plistextra){
...
+    if(length(geom.name.list) > 0){
+      len.list <- length(geom.list)
+      geom.name.list <- geom.name.list[1:len.list]
+    }
+    # only use geoms not ignored earlier
+    geom.name.list <- geom.name.list[!is.ignored]
+    
+    ## guides_geom() earlier returned geom names that are not
+    ## reflected by the ggtype function
+    ## convert these to get proper geom names for legends
```
faizan-khan-iit: `guides_geom()` now returns a `draw_key()` method instead of a geom name. So I had to get the geom names from the layers. But the geom name returned by the `guides_geom()` is not the same as the one in the layers, hence I had to resort to this for loop. Will have to keep updating for more cases unless we find a better solution to this.

faizan-khan-iit:
Some progress in getting the legends to work.

This basic example works well (See here: [with ggplot2 v1.0.1](http://bl.ocks.org/faizan-khan-iit/raw/e41d81592464eae21acb067954b178b0/), [with ggplot2 v2.1.0](http://bl.ocks.org/faizan-khan-iit/raw/0a936cddc98f42aa1aecaa40055d111c/))
```
set.seed(1410)
data("diamonds")
dsmall <- diamonds[sample(nrow(diamonds), 100), ]

w1 <- qplot(carat, price, data = dsmall, colour = color)
w2 <- qplot(carat, price, data = dsmall, shape = cut)
w3 <- qplot(carat, data = dsmall, geom = "density", colour = color)
w4 <- qplot(carat, data = dsmall, geom = "histogram", fill = color)

w <- list(w1=w1, w2=w2, w3=w3, w4=w4)
```
The 3rd graph renders a little differently. Will have to look into that.
The changes seem fine on the ~20 examples that I tested. I will be testing the changes on the legends with a few more extensive examples.

@kferris10 @cpsievert Then I guess its okay to continue with the changes I made, since they seem to be working fine.

faizan-khan-iit:
@tdhock I had a little doubt. As you may see in the latest [wercker build](https://app.wercker.com/#buildstep/57606a23a4b49c7d1e021052), the tests for `vjust` and `hjust` fail because they are now passed along with the data with a default value of 0.5. Hence [this condition] (https://github.com/tdhock/animint/blob/610ef59723fb9394d1b50ad31412a6562cd7424d/R/animint.R#L580-#L581) always evaluates to `TRUE`. So we get a warning every time we use `geom_text` without specifying `vjust`, which messes up the tests. 

I wanted to ask whether I should edit the compiler to ignore the `vjust` every time? If not, we will get the same warning every time we use `geom_text` without `vjust=0`, like the 23 warnings in the build.

tdhock:
is it possible to test if the user has manually specified a vjust other than 0? (in the params or as an aes in the data)

the idea was to tell the user that animint does not support vjust values other than 0.

if it is too complicated you can just remove the warning and test.

anyways we should eventually implement vjust, and you can do that if you want. if you do, please add tests for vjust=0, 0.5, 1. you can use the `measureText` function in javascript https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L121 to get the height of the text element in pixels, and from there you can compute the y position offset.

tests/testthat/test-renderer2-showSelected-same.R
```diff
@@ -25,8 +25,10 @@ test_that("redundant aes not saved to tsv", {
   geom.tsv <- Sys.glob(
     file.path("animint-htmltest", "geom1_point_points_*.tsv"))
   expect_equal(length(geom.tsv), 1)
-  expected.names <- sort(c("x", "y", ss.vec))
-  tsv.data <- read.table(geom.tsv, header=TRUE)
+  expected.names <- sort(c("x", "y", ss.vec,
+                           "alpha", "colour", "fill",
+                           "shape", "size", "stroke"))
```
tdhock: Hey @faizan-khan-iit can you please revert this commit? In this test case, the tsv file should definitely NOT contains alpha/colour/fill/shape/size/stroke columns.

If these columns are output in the built data from the new ggplot2, then we need to write code that analyzes which columns are constant, and then removes those columns and stores those data as params (to reduce tsv file sizes).

tests/testthat/test-renderer1-geom-abline.R
```diff
@@ -6,10 +6,11 @@ p <- qplot(wt, mpg, data = mtcars) +
 info <- animint2HTML(list(p = p))
 
 tsv.file <- file.path("animint-htmltest", "geom2_abline_p_chunk1.tsv")
-tsv.data <- read.table(tsv.file, header=TRUE)
+tsv.data <- read.table(tsv.file, header=TRUE, comment.char = "")
 
 test_that("columns of abline tsv", {
-  expected.names <- sort(c("PANEL", "x", "xend", "y", "yend"))
+  expected.names <- sort(c("PANEL", "x", "xend", "y", "yend",
+                           "colour", "size", "linetype", "alpha"))
```
tdhock: @faizan-khan-iit again can you please revert this test case? The tsv file should not contain any constant columns.

tests/testthat/test-renderer1-WorldBank-NA.R
```diff
@@ -67,7 +67,7 @@ test_that("common chunk contains expected columns", {
   expected.cols <-
     c("ymin", "ymax", "xmin", "fill", "key",
       "clickSelects", "showSelectedlegendfill",
-      "group")
+      "group", "size", "colour", "linetype")
```
tdhock: @faizan-khan-iit  same here please revert this test case.

tests/testthat/test-renderer2-showSelected-same.R
faizan-khan-iit: @tdhock Thanks for bringing this up! It was one of the matters I wanted to discuss. I'll see how I can implement this with the new ggplot2 internals.

faizan-khan-iit:
@tdhock Could you also take a look at this - https://github.com/tdhock/animint/pull/135/commits/588e92a58e65412ba46181dceaad1bf5f55e26d7

ggplot2 now takes [a single argument for `labeller`](https://github.com/tdhock/animint/pull/135/files#diff-77f1d715bd783e7e9fd141e37d4d134aL37), so the earlier tests failed as they used both `val` and `var`. ggplot2 returns an error for extra arguments.

So should we support the previous method of passing both arguments or move ahead with the current ggplot2 syntax? Currently its up to user to define the `labeller` accordingly.

faizan-khan-iit:
@tdhock We are passing almost all the tests now. 

There are two points worth noticing:

1. The [save-separate-chunks tests fails](https://travis-ci.org/tdhock/animint/jobs/138855289#L2939) because we have an extra column `anchor` in the data. As a result `is.common` vector [here](https://github.com/tdhock/animint/blob/a7651eff5ab80f7333481fc6cc55fe0b34499314/R/animint.R#L1097) has sum>=2. So a common chunk file gets saved. The file only has two columns `anchor` and `group` with a single entries `middle` and `1` respectively. I will take a look into fixing this if this doesn't get too complicated.

2. The code has slowed down a bit, which leads to the failure of the wercker build. I will run the profiler to get a better idea of what is slowing down the execution, once I get all the tests to pass.

R/animint.R
```diff
@@ -130,11 +133,12 @@ parsePlot <- function(meta){
     ## Data now contains redundant columns with fill, alpha, colour etc.
     ## Remove from data if they have a single unique value and
     ## are NOT used in mapping to reduce tsv file size
+    redundant.cols <- c("fill", "shape", "color", "colour", "stroke", "alpha",
+                        "linetype", "size", "angle", "vjust", "family",
+                        "fontface", "lineheight", "anchor")
```
tdhock: @faizan-khan-iit is there a reason why we need to list all the column names in the R code, rather than just detecting them based on the data?

faizan-khan-iit: @tdhock I just thought it would be safer to remove the ones listed here. Ideally, it should work without specifying the columns too. But I specified them to be on the safe side.

R/animint.R
```diff
@@ -696,9 +733,51 @@ saveLayer <- function(l, d, meta){
...
+        g.data.i[[col.name]] <- scales::rescale(g.data.i[[col.name]], 
+                                                0:1, ranges.i$x.range)
+      } else if(grepl("^y", col.name) || grepl("^Y", col.name)){
+        g.data.i[[col.name]] <- scales::rescale(g.data.i[[col.name]], 
+                                                0:1, ranges.i$y.range)
+      }
+    }
+    g.data.i
+  }
+
```
faizan-khan-iit: Do we ever get column names starting with uppercase X/Y like `XEND`/`YINTERCEPT` etc? I don't think so. In that case the second condition in the `if` statements are redundant and may be removed. In here and some other places that I have used this...

R/animint.R
```diff
@@ -1644,6 +1746,36 @@ getLegendList <- function(plistextra){
...
+        geom.legend.list[[i]] <- substr(geom.legend.list[[i]], 1,
+                                        nchar(geom.legend.list[[i]])-1)
+      }else if(geom.legend.list[[i]] %in% c("segments")){
+        geom.legend.list[[i]] <- "path"
+      }else if(geom.legend.list[[i]] %in% c("rect")){
+        geom.legend.list[[i]] <- "polygon"
+      }
+    }
+    
+    gdefs[[leg]]$geom.legend.list <- geom.legend.list
```
faizan-khan-iit: I have to evaluate the `draw_key` function here to produce a `grid` object and take the geom name from there. It seems like too much computation when we just want the geom name to be used for getting legends.

I tried [using another approach](https://github.com/tdhock/animint/pull/135/files/adac7130a34196d6414c6c1411124c2bed432f09..99509efd12d562d78f6d8eec35b6f7fbfaa245b0#r65360704) before but it gets complicated for multiple geoms.

tests/testthat/helper-functions.R
```diff
@@ -195,24 +195,13 @@ getStyleValue <- function(html, xpath, style.name) {
...
-             " warning", s, ": ",
-             paste(warnings, collapse=", ")),
-      "no warnings given"
-    )
-  }
-}
-expect_no_warning <- function(object, ..., info=NULL, label=NULL){
-  if (is.null(label)) {
-    label <- testthat:::find_expr("object")
-  }
```
faizan-khan-iit: `testthat` has changed too and `testthat:::find_expr` now [throws an error](https://app.wercker.com/#buildstep/5762874ea79f5ea83002d120). Had to try a different approach for `expect_no_warning`

R/geoms.R
```diff
@@ -8,30 +8,62 @@
 #' @export
 #' @import proto
 #' @example inst/examples/breakpoints.R
```
faizan-khan-iit: I don't think we need `proto` now. As a sidenote, do we use `proto` somewhere else in the code too? Otherwise we may edit the DESCRIPTION file.

R/animint.R
tdhock: yeah I think you can get rid of the X/Y code

by the way you could have used the pattern `"^[xX]"` or the `ignore.case=TRUE` argument

R/geoms.R
tdhock: yeah we can remove `proto` (new ggplot2 uses `ggproto` instead)

tests/testthat/helper-functions.R
tdhock: that is probably one reason why there is a conflict with master --- we have already merged changes to `expect_no_warning` https://github.com/tdhock/animint/blob/master/tests/testthat/helper-functions.R#L228

I will pull changes from master into this branch and try to resolve the conflicts.

R/animint.R
tdhock: the more R-ish way to do this would be using data and vector operations rather than a for loop, e.g.

```r
convert.names.vec <- c(segments="path", rect="polygon")
geom.legend.list[geom.legend.list %in% names(convert.names.vec)] <- convert.names.vec[geom.legend.list]
```

the advantage is that the mapping is defined in data so the code is clearer and easier to understand

tdhock:
Hey @faizan-khan-iit I pulled changes from master and fixed the conflicts, but now some tests are failing. Can you please fix them before we merge this branch with master?

faizan-khan-iit:
@tdhock Sure. Let me take a look.

faizan-khan-iit:
@tdhock I've got some minor doubts:

1. I updated the documentation for `geom_tallrect` and removed `proto` from `DESCRIPTION` file. Upon running `devtools::document` I noticed that an extra file for `getCommonChunk` was generated. Did we exclude that before? (if its OK, I will push these changes)
2. Also its OK if I edit the `NAMESPACE` file by hand to exclude `proto` from imports, right? I asked because `roxygen` won't overwrite it.
3. Since merging of conflicts, the wercker and travis builds fail due to an error. Should I revert the changes to the `travis.yml` and `wercker.yml` files? Or could you take a look?

P.S. All tests are passing locally now.

tdhock:
great to hear that all tests are passing locally!

1. please `git add` any new Rd files (we probably just forgot to add it before)
2.yes it is ok to edit the NAMESPACE
3. yes go ahead and edit the wercker and travis files

faizan-khan-iit:
@tdhock Anything else we need to work on here? Coverage is almost nil, for example. 

If not we could just bump up the version and update the NEWS for this PR, and we will be ready to merge this.

tdhock:
tests pass, which is good. but there are some warnings that worry me.

```
> filter <- "renderer1"
> tests_run(filter=filter)
[1] "axis angle rotate"
[1] 0.75
[1] "One character value"
[1] "coord"
[1] "facet-coord"
[1] "facet lines"
[1] "facet_grid(space=\"free\")"
[1] "facet-strips"
[1] "facet-trivial"
[1] "geom_abline"
[1] "global variables"
[1] "hline"
[1] "hjust text anchor"
[1] 0.8
[1] 1.5
[1] 0.5 1.0
[1] 1
[1] 0.5
[1] 0.7
[1] "aes(href)"
[1] "HTML page title"
[1] "Interactive facets"
[1] "knitting multiple animint plots in a single Rmd"


processing file: index.Rmd
  |......                                                           |  10%
  ordinary text without R code

  |.............                                                    |  20%
label: safety (with options) 
List of 1
 $ echo: logi FALSE

  |....................                                             |  30%
  ordinary text without R code

  |..........................                                       |  40%
label: plot-1-top
  |................................                                 |  50%
  ordinary text without R code

  |.......................................                          |  60%
label: plot.1.bottom
  |..............................................                   |  70%
  ordinary text without R code

  |....................................................             |  80%
label: boring
  |..........................................................       |  90%
  ordinary text without R code

  |.................................................................| 100%
label: breakpoints

output file: index.knit.md

/home/tdhock/bin/pandoc +RTS -K512m -RTS index.utf8.md --to html --from markdown+autolink_bare_uris+ascii_identifiers+tex_math_single_backslash --output index.html --smart --email-obfuscation none --standalone --section-divs --template /home/tdhock/lib/R/library/rmarkdown/rmd/h/default.html --variable 'theme:bootstrap' --include-in-header /tmp/RtmpxxIl1e/rmarkdown-strb6228cebd8a.html --mathjax --variable 'mathjax-url:https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML' --no-highlight --variable highlightjs=index_files/highlight --variable navigationjs=index_files/navigation-1.0 

Output created: index.html
[1] "Labels"
[1] "legends"
[1] "linetype"
[1] "malaria data viz"
[1] "many facets"
[1] "Panel background"
[1] "params"
[1] "PeakConsistency"
[1] "rect size"
[1] "Segment size"
[1] "Text"
[1] "aes(tooltip)"
[1] "variable value"
[1] "WorldBank-NA"
testthat results ================================================================
OK: 419 SKIPPED: 0 FAILED: 0
There were 18 warnings (use warnings() to see them)
> warnings()
Warning messages:
1: In animint2dir(plotList, out.dir = "animint-htmltest",  ... :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=year), problem: geom6_text_ts
2: In geom.legend.list[names.to.change] <- convert.names.list[unlist(geom.legend.list)] :
  number of items to replace is not a multiple of replacement length
3: Using size for a discrete variable is not advised.
4: Using size for a discrete variable is not advised.
5: Using size for a discrete variable is not advised.
6: Using size for a discrete variable is not advised.
7: Using size for a discrete variable is not advised.
8: Using size for a discrete variable is not advised.
9: Using size for a discrete variable is not advised.
10: Using size for a discrete variable is not advised.
11: `show_guide` has been deprecated. Please use `show.legend` instead.
12: `show_guide` has been deprecated. Please use `show.legend` instead.
13: In animint2dir(plotList, out.dir = "animint-htmltest",  ... :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=year), problem: geom3_rect_scatter
14: In animint2dir(plotList, out.dir = "animint-htmltest",  ... :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=year), problem: geom4_rect_scatter
15: In animint2dir(plotList, out.dir = "animint-htmltest",  ... :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=year), problem: geom5_text_scatter
16: In animint2dir(plotList, out.dir = "animint-htmltest",  ... :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=year), problem: geom8_bar_bar
17: In animint2dir(plotList, out.dir = "animint-htmltest",  ... :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=year), problem: geom3_text_scatter
18: In animint2dir(plotList, out.dir = "animint-htmltest",  ... :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=year), problem: geom3_text_scatter
> 
```

In particular this one involves the new code and is probably a bug. Can you please investigate and fix?

```
> tests_run(filter="malaria")
[1] "malaria data viz"
testthat results ================================================================
OK: 8 SKIPPED: 0 FAILED: 0
Warning message:
In geom.legend.list[names.to.change] <- convert.names.list[unlist(geom.legend.list)] :
  number of items to replace is not a multiple of replacement length
> 
```

tdhock:
you are right, coverage has been down near zero since the covr package changed how they compute the coverage. I think it has something to do with the fact that we run tests in 4 different builds on the travis build matrix. Maybe there is some way to get it to work on wercker? Or to compute the coverage differently? We may want to file an issue with the covr package, but for now I would say it is not top priority (we can do that in a later PR).

faizan-khan-iit:
Regarding the warnings:

- `aes(key)` warning: Still working on that
- geom.legend.list warning: Fixed
- `size` for discrete variable warning: Now ggplot2 warns us about this every time we produce a discrete plot with `size` aesthetic, e.g. (from test-renderer1-panels.R):

```
library(ggplot2)
p1 <- ggplot() +
  geom_point(aes(Sepal.Length, Sepal.Width,
                 colour = Species, size = Species), data = iris)
p1
```
```
Warning message:
Using size for a discrete variable is not advised. 
```
- `show_guide` warning: `show_guide` has been renamed to `show.legend` - Fixed.

faizan-khan-iit:
@tdhock I was looking into the `aes(key)` warning, and I've to say that I can't think of anything that will solve this. The best I could come up with was that in some cases, we may not need smooth transitions at all (like the [lilac chaser visual illusion](http://bl.ocks.org/faizan-khan-iit/035cb159f23f3469abef710ec2ce80ae)). In which case I could check if `duration` is set to zero and skip the warning. Although this won't avoid all the warnings, it will work when the user explicitly wants instantaneous transitions.

cpsievert:
Nice work @faizan-khan-iit! Unfortunately, when I run tests I run into some errors...any ideas?

https://gist.github.com/cpsievert/6a3dfc051a44b5c1e883c569ba9ee88f

faizan-khan-iit:
@cpsievert Thanks!

That is probably because the ggplot v2.1.0 does not allow extra parameters now. So `chunk_vars` creates an error. So we had to submit a PR to get this to work. You may [see here](https://github.com/hadley/ggplot2/pull/1649).

Did you try installing animint from this branch?
```
devtools::install_github("tdhock/animint", ref = "new-ggplot2")
```
This should install the ggplot version that works with animint.

Or you could try installing `ggplot2` from my `validate-params` branch directly:
```
devtools::install_github("faizan-khan-iit/ggplot2", ref="validate-params")
```

tdhock:
I increased the version number and will merge with master. I am excited to be able to use the new version of the code for my tutorial at useR tomorrow, and I will tell the audience about your contribution Faizan. Good job!

faizan-khan-iit:
@tdhock Thanks a ton!

kferris10:
Awesome work @faizan-khan-iit !  Really excited that this is done.

caijun:
Good job @faizan-khan-iit !

faizan-khan-iit:
@kferris10 Thanks Kevin! Your initial commits helped me a lot in pinpointing the major changes I needed to make.
@caijun Thanks!

PR #138, opened 2016-03-08
print.animint

tdhock:
say a new user is reading a tutorial written on an Rmd document, where the animint is rendered via `structure(viz, class="animint")`

they will probably just want to copy that code and paste it into their terminal to see the interactive data viz

to accomplish that, we can define the `print.animint` method which just calls `animint2dir`


tdhock:
TODO write a test!

PR #140, opened 2016-03-13
fill transparent fails

tdhock:
a small bugfix for `geom_point(aes(color), fill=NA)` which should render colored stroke and transparent fill.


tdhock:
problem with test infrastructure: rgba(0,0,0,0) should be treated as equal to "transparent" -- need to edit ensure_rgb in helper-functions.R

PR #146, opened 2016-04-01
warn for geoms with showSelected=duration var but no key

tdhock:
if we use smooth transitions a key should be specified to ensure that those transitions are meaningful. the compiler can and should warn the user if the key should be specified, but is not.


tdhock:
aes(href) is working for geom_point but not for geom_polygon. The `a` elements are rendered but they do not have an `href` attribute. this is probably an easy fix for the renderer.

tdhock:
TODO add a test that captures #147

Issue #155, opened 2016-06-28
Points don't render when both href and tooltip aesthetics are used


analyticalmonk:
For example,
```
library(animint)
data(WorldBank, package = "animint")

ex_plot <- ggplot() +
    geom_point(aes(fertility.rate, life.expectancy, color = region,
                   tooltip = country, href = "https://github.com"),
               data = WorldBank[WorldBank$year == 1975, ])

plot_list <- list(ex = ex_plot)
structure(animint, class = "animint")
```
The above code results in an animint plot without the required points.
![animint_hreftooltipissue2](https://cloud.githubusercontent.com/assets/4419448/16407237/50bfeafa-3d30-11e6-9f23-f915d6ffb16c.png)

Removing either of the two (tooltip or href) provides the expected plot.
![animint_tooltiphrefissue](https://cloud.githubusercontent.com/assets/4419448/16407259/6624dbb2-3d30-11e6-8c22-ab79ffe22ec8.png)

@faizan-khan-iit Please look into this.

_Note: As discussed in animint issue #100 and Rperform issue [#15](https://github.com/analyticalmonk/Rperform/issues/15)_

faizan-khan-iit:
@analyticalmonk Thanks for pointing this out Akash. I will take a look into this ASAP.

faizan-khan-iit:
@analyticalmonk Hey Akash, the problem with using both `href` and `tooltip` has been fixed.

Here's your example - http://bl.ocks.org/faizan-khan-iit/raw/d8671bf1b2450213348b39373d7a31e1/

Please inform if you have any problem working with this.
Cheers!

faizan-khan-iit:
Closed by #156

Issue #100, opened 2015-07-19
Issue with Tooltips


kferris10:
Does anybody know why there might be problems with the tooltips at the end of [this post](http://kferris10.github.io/AnimintBlog/Hour-by-Hour-Visualization-of-the-Ridge-Hike/)?

If I take out the `aes(href = links)` argument, the points render just fine.

kferris10:
It looks like the points are rendering as `<a>` elements instead of circles

![screenshot](https://cloud.githubusercontent.com/assets/6612411/8764089/fa815fe8-2d7c-11e5-82d4-72cdf39d89b6.PNG)

cpsievert:
Not sure why this happening, but it certainly looks like a bug.

faizan-khan-iit:
Closed by #156

PR #156, opened 2016-06-29
Bug when trying to use both 'href' and 'tooltip'

faizan-khan-iit:
This PR covers Animint issues #100 and #155


inst/htmljs/animint.js
```diff
@@ -1696,8 +1696,7 @@ var animint = function (to_select, json_file) {
 	  return d["clickSelects.variable"] + " " + d["clickSelects.value"];
 	};
       }
-      elements.text("")
-        .append("svg:title")
```
faizan-khan-iit: What is `elements.text("")` doing here? Removing it seems to solve the problem...

faizan-khan-iit:
After this minor change, everything seems to work fine.

Here is the example from #155 - [Worldbank example](http://bl.ocks.org/faizan-khan-iit/449b84b2e6b55ca5e39ea9aaad97e27e)
And here is the example from #100 - [Visualization of Ridge Hike](http://bl.ocks.org/faizan-khan-iit/0b6f5c86443774f7c6e0d83708f9ff06)

But the wercker build is stuck for some reason. Maybe ignore it since the Travis build is passing?

@tdhock @cpsievert @kferris10 Could one of you please confirm?

tdhock:
Thanks for working on this Faizan. Before starting to code various solutions, can you please add at least one test case that fails?

tdhock:
Also please add a test case for the bug in your approach that you found.

faizan-khan-iit:
@tdhock Sorry about that. I found out about the bug just before I had to leave so I thought to put in a comment so that you people do not waste your time.

So in the initial bug that we were trying to fix, `<circle>`  was getting overwritten by `elements.text("")` when using tooltip with href since the both `circle` and `title` were children of anchor tag `a`. So I thought removing `elements.text("")` should do the job. Which led to [this viz](http://bl.ocks.org/faizan-khan-iit/449b84b2e6b55ca5e39ea9aaad97e27e), as I mentioned above.

Initially it works fine, but while interacting with the points, the code keeps appending `title` elements one after the other, which is the bug I noticed. If you navigate to the above viz, initially all is good. Notice the Canada and US points in blue in the top left of the plot. But once you hide all the points excluding North America, you will see that their tooltip labels are now Armenia and Albania (that's what I get). The titles get updated every time I hide or unhide elements, and the topmost value is displayed every time.

Hence I currently remove `<tilte>` every time an update takes place. This results in [this viz](http://bl.ocks.org/faizan-khan-iit/6a483d9bd03f2c939ceb11ecc294d22c), which seems fine for all the testing I have put it through.

Hopefully this explains the changes and tests that I made.

Still the tooltip test gets stuck somewhere. I will see what is causing this

tdhock:
phantomjs has problems when there are several elements with the same ID -- is that the problem?

faizan-khan-iit:
@tdhock I don't think so. The HTML is quite straight-forward and there are no ID clashes. I specify the IDs to be clicked [here](https://github.com/tdhock/animint/pull/156/files#diff-917aab0dfef94d843e17604a14008ed2R139).

I tried to find out where the execution gets stuck, and I found out it gets stuck here, at [the first getHTML() call](https://github.com/tdhock/animint/pull/156/files#diff-917aab0dfef94d843e17604a14008ed2R149).

Should I write a different test if this doesn't work?

tdhock:
try removing non-alphanumeric characters from the ids (or convert them to underscores). I have a feeling that `(&)` characters are the problem.

faizan-khan-iit:
@tdhock Great observation! Once I filtered out the non-alphanumeric chars, the tests pass locally using `tests_run(filter="tooltip")`. 

But the bad news is that they get stuck when running `tests_run(filter="renderer1")`, like in the builds. 

I separated the tests which are causing the problem. Probably ignore these for now? We could check this separately using firefox locally...

tdhock:
Try renaming that test file to `test-renderer4-tooltip.R` and then adding `TEST_SUITE=renderer4` to the travis and wercker builds.

faizan-khan-iit:
travis seems fine with the changes but looks like wercker is still stuck with the new tests.

faizan-khan-iit:
@tdhock Could you please restart again. It failed before starting the tests...

tdhock:
looks like there are still problems. can you double-check everything on that page to make sure there are no duplicate id values, and no id values with non-alphanumeric characters? you may even want to try deleting `_-` characters to see if that helps, and make sure that the id does not start with a number. Not sure why all of this is necessary, I guess it is some bug in phantomjs...

faizan-khan-iit:
@tdhock Looks like wecker build failed again without starting the tests..

I cut the region names to the first three letters. So these are the html id's we are working with now https://travis-ci.org/tdhock/animint/jobs/142332160#L2663

I will remove these once we complete the wercker build. Should I try and remove the rest of the underscores too? I would probably need to change the JS code, which may cause unexpected problems in the other tests.

Also there are no conflicts whatsoever in the rendered HTML. The only point worth noting is that the `svg` and `label` elements under these IDs mentioned have IDs with "_svg" and "_label" appended to them. Eg.
Legend ID: plot_ex_region_variable_eas
SVG ID: plot_ex_region_variable_eas_svg
Label ID: plot_ex_region_variable_eas_label

But that should not be problem since there are no clashes.

tdhock:
@cpsievert can you please give @faizan-khan-iit permissions to restart your wercker build?

faizan I don't understand why the wercker tests fail... can you try simplifying the test or using a different data set?

also while we debug this, feel free to start working on something else in another branch...

faizan-khan-iit:
@cpsievert Thanks!

@tdhock Looks like the new tests work well with wercker. As for the older tests, I have not deleted those yet, but removed them from the test suites. If you approve of this, I could update the NEWS and DESCRIPTION.

A minor note: I had to install ggplot2 separately for wercker. Don't know why but it was not installing ggplot2 from our PR...

tdhock:
+1 looks good to me, go ahead and increase the version in NEWS and DESCRIPTION, then you can merge with master

Issue #127, opened 2015-12-20
aes(stroke) -> css stroke-width


tdhock:
This is a new aesthetic in ggplot2 v2.0 that we can use to render the css stroke-width property

```
geom_point() gains a stroke aesthetic which controls
 the border width of shapes 21-25 (#1133, @SeySayux).
 size and stroke are additive so a point with size = 5 and 
stroke = 5 will have a diameter of 10mm. (#1142)
```

https://github.com/hadley/ggplot2/blob/master/NEWS.md#ggplot2-200

tdhock:
@faizan-khan-iit I don't think this is a super important issue, but it should be an easy fix, now that you got animint working with the new ggplot2 >= 2.0.

faizan-khan-iit:
Closed by #157

PR #157, opened 2016-07-08
Stroke width aesthetic for geom_point

faizan-khan-iit:
This PR covers issue #127

### DIFF
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1082,7 +1082,7 @@ var animint = function (to_select, json_file) {
     };
     
     // stroke_width for geom_point
-    var stroke_width = 0.5;  // by default ggplot2 has 0.5
+    var stroke_width = 1;  // by default ggplot2 has 0.5, animint has 1
     if (g_info.params.hasOwnProperty("stroke")) {
       stroke_width = g_info.params.stroke;
     }
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -1080,7 +1080,19 @@ var animint = function (to_select, json_file) {
       }
       return size;
     };
-
+    
+    // stroke_width for geom_point
+    var stroke_width = 0.5;  // by default ggplot2 has 0.5
+    if (g_info.params.hasOwnProperty("stroke")) {
+      stroke_width = g_info.params.stroke;
+    }
+    var get_stroke_width = function (d) {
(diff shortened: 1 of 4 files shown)


faizan-khan-iit:
Some comments here:

With the current implementation, the animint and the ggplot will look a little different.
As quoted in the issue #127 -

> size and stroke are additive so a point with `size = 5` and `stroke = 5` will have a diameter of 10mm

But while using SVG, strokes are aligned to the center of each shape’s edge. That is, the stroke is neither fully inside nor outside the object, but is both half in and half out. So `size=5` and `stroke=5` gives a point with diameter 7.5 mm, with a stroke of width 5mm.

Do we need to fix that?

inst/htmljs/animint.js
```diff
@@ -1080,7 +1080,19 @@ var animint = function (to_select, json_file) {
       }
       return size;
     };
-
+    
+    // stroke_width for geom_point
+    var stroke_width = 0.5;  // by default ggplot2 has 0.5
```
tdhock: the previous default in animints seems to be stroke_width=1 so can we please keep that?
screenshot-stroke-width

tests/testthat/test-renderer4-geom-point-stroke.R
```diff
@@ -0,0 +1,57 @@
...
+
+p2 <- ggplot(mtcars, aes(wt, mpg, stroke=cyl)) +
+  geom_point(shape = 21, colour = "black", fill = "white", size = 5)
+
+viz <- list(p1=p1, p2=p2)
+info <- animint2HTML(viz)
+
+test_that("points are rendered with stroke-width", {
+  point_nodes <-
+    getNodeSet(info$html, '//g[@class="geom1_point_p1"]//circle')
```
tdhock: please use `getStyleValue(, , "stroke-width")`

tdhock:
good job on this first implementation!

It would be nice if the animint looked exactly like the ggplot, but in this situation it seems too complicated to achieve. so no need to fix that.

for the tests, please add `expect_equal(stroke_width_from_html, stroke_from_R_code)`

faizan-khan-iit:
@tdhock Thanks for the tip! Using that initially would have saved me some effort..

tdhock:
looks good to me +1

go ahead and edit DESCRIPTION/NEWS, then merge with master

PR #158, opened 2016-07-12
Update axes/legends after displaying current data subset

faizan-khan-iit:
This PR covers the updating of axes while displaying a different data subset in a plot.
Here is an example of what we aim to achieve: https://dc-js.github.io/dc.js/crime/index.html

A new option `theme_animint(update_axes = c("x", "y"))` will be implemented for this purpose


faizan-khan-iit:
@tdhock A very raw and basic example: http://bl.ocks.org/faizan-khan-iit/raw/810a77ee29dcc110392305a80df86279/

This is a very basic implementation and if you are OK with it, I will continue to work on this. There are quite a lot of problems currently as I have only updated the scales for the plot, but the data is getting re-scaled as expected so I guess this approach should be OK.

Any major changes that you would like in this?

inst/htmljs/animint.js
```diff
@@ -1766,6 +1771,17 @@ var animint = function (to_select, json_file) {
...
       selectized_array[v_name].setValue(selected_ids, true);
     }
+
+    // For each updated geom, check if the axes of the plot need to be
+    // updated and update them
+    s_info.update.forEach(function(g_name){
+      var plot_name = g_name.split("_").pop();
+      var axes = Plots[plot_name]["options"]["update_axes"];
+      if(axes != null){
+        update_scales(plot_name, g_name, axes, v_name, value);
```
tdhock: hey thanks for sharing this first implementation

I find it strange that `update_scales` takes `g_name` as a parameter. Why do you need call it for every geom?

I would think that after a click to change a selection variables, we only need to update each axis once.

faizan-khan-iit: That is what I wanted to implement too at first, because its simpler. But there might be problems with that, since we are updating the scale itself. 

For example [in this example](http://bl.ocks.org/faizan-khan-iit/810a77ee29dcc110392305a80df86279) in the plot.json file under `axis_ranges` we have 2 geoms: `text` and `bar` (you could search for 'axis_ranges' in the page). Since we will be updating the scale for the plot every time before drawing the geom, the text may be rendered outside the boundaries of the plot. Click on Texas in the viz and you can barely see the state at the top. It will need to be fixed eventually so I thought we will have to update all the geoms.

Any ideas to implement this in a different manner?

tdhock: I would first update the scales, and then update the geoms (which will probably use the scales to compute their positions)

faizan-khan-iit: @tdhock Sorry but could you please explain a bit? 
Currently I cycle through all the geoms in `s_info.update`. If the geom belongs to a plot that needs to have its axis updated, I take the appropriate ranges, and update the scales of the plot accordingly.  For eg, the geom_text will not have any effect on scales as it does not have a "state" selector, so we get `undefined` [here](https://github.com/tdhock/animint/pull/158/commits/a7370f4fb5a526d247a6ac7c962962ea20c3680a#diff-bfbd034b190fd160d92db9a24ea2a3c7R1733).
Then the `update_geom` function gets called as usual.

As far as I can understand, you are suggesting that I compute a single range value for a data subset for the whole plot in the compiler, which can be used to update the plot scales every time any geom belonging to it is updated. Is that correct?

tdhock: right! there is no need to cycle through all those geoms in the renderer, when we could just do it once at compile time

faizan-khan-iit:
@tdhock I removed using the `g_name` param while using `update_scales`. Now we only save a single subset range array for a selector in the `plot.json` file. But I still have some doubts that I would like to clear.

In the example that I am working on, we have two geoms in the plot that need to be updated. `geom_bar`  is used to calculate the subset domains currently, and these get saved to the `plot.json` file. The problem is `geom_text`, which has a fixed domain (`[200,200]` in the example) at the top of the plot. How do I change its domain every time I update the plot, so that it gets displayed? I could somehow compute its relative position in the compiler, but I would still need to save it separately from the other domain. One possible solution is to check for the maximum of both domains and use that in our scales.

P.S. just realized it would be better to call these `domains` instead of `ranges`, so will edit the code accordingly to avoid confusion in renderer code.

tdhock:
let's clarify the terms

domain is the input to a function (here, a scale). the domain is the input data values, in units specified by the data set. Typically the units of the domain are specified in the axis label. For your example the domain is the number of tornadoes in a state.

range is the output of the scale function. the range in animint is (1) for axes, the pixel values at which the geoms will be drawn, or (2) the visual properties of the geoms drawn in the legend (color, linetype, etc).

to me, it does not make sense to talk about the domain and range for specific geoms. each plot has at least two axes (more if there are several facets/panels), each of which has a scale with a domain and range. Each plot also may have legends, each of which have a scale with a domain and range. Each of these scales should be updated when the showSelected subset changes.

so please do not compute domains for each geom. Do it for each panel and showSelected subset (take the min/max across all geoms that appear in that panel, for that showSelected subset).

faizan-khan-iit:
@tdhock Thanks for the thorough explanation!

Sorry if my last comment was a bit confusing. What I actually meant to ask was *'what to do with constant ranges?'*. I'll explain:

In the latest commit, I cycle through all the geoms in the plot with update_axes option, take the min/max amongst the ranges that have a showSelected subset for the selector and save it for the renderer. Hence a single unique domain for a plot gets saved. The result is [this viz](http://bl.ocks.org/faizan-khan-iit/359faf916b49ae537e828a502036ed4a). Upon interacting with the viz, you will notice that the plot actually gets zoomed with every subset, but not adequately. The problem lies in the [ranges stored here](http://bl.ocks.org/faizan-khan-iit/359faf916b49ae537e828a502036ed4a#plot.json), which always have a topmost value of `200`, due to `geom_text`. For another example [consider this viz](http://bl.ocks.org/faizan-khan-iit/8d641ae6b5af9bb68bb127e3651a1f62), where I have made this minor change to the previous code:
```
ts=ggplot()+
         # Use y parameter as 100 instead of 200
         make_text(UStornadoes, 1980, 100, "state")+
```
Now if you compare them by clicking Texas(TX), Okhlahoma(OK) & Kansas(KS) (I hope I got the names right) in both the viz, you will see that in the second example, all subsets with max range > 100 are getting zoomed in properly. I hope this clarifies the confusion I caused, and also explains the problem.

Also regarding the domains/ranges, I do get the difference between them. I was just suggesting that it would be better if I saved the info in the plot.json file as `axis_domains` instead of `axis_ranges`, so that  others may not get confused in the future.

tdhock:
ok thanks for clarifying

I guess it is not really a problem with the animint compiler or renderer -- it is a problematic definition of the geom_text y value by the designer. in other words, the designer should specify a different y value for the text for every showSelected subset. say y=max(tornadoes) for each subset and then that would be fine, right?

also not exactly sure what you meant by "take the min/max amongst the ranges that have a showSelected subset" but the correct behavior is to take the min/max of all data that will appear for a particular showSelected value. another way to think of it would be: use the same scales as if you did `facet_grid(state ~ ., scales="free")` in a static ggplot.

about the names, please change the names in the plot.json file and the code so that it is easier to understand!

faizan-khan-iit:
@tdhock Exactly! I spent quite a lot of time worrying about that, but I guess leaving it to the user is the best solution. Then I do not have to worry about that anymore!
The current implementation seems to fulfill all the criteria nicely, so I will continue with this.

faizan-khan-iit:
Some more progress. This looks better than before: http://bl.ocks.org/faizan-khan-iit/raw/e05d84797bb0417137f231b07895cc90/

faizan-khan-iit:
Current progress: http://bl.ocks.org/faizan-khan-iit/raw/5ea2f8cdaa168959a5d6e49f6027b9ba/
Still working on handling axis ticks and grid lines.

tdhock:
I will try to create a couple of realistic test cases.
- genomic data with x axis that updates https://github.com/tdhock/animint-examples/blob/master/examples/chip.seq.R
- facet_grid(scales="free") should take scales across all relevant panels

tdhock:
two kinds of tests would be useful:
- axis ticks and grid lines. measure the space between them in pixels using https://github.com/tdhock/animint/blob/master/tests/testthat/test-renderer1-facet-space.R#L71
  - when no axis zoom is specified, there distance between ticks/gridlines should be the same before and after a click.
  - when axis zoom is specified, there should be different ticks, and/or the distance between ticks should be different before and after a click.
- data limits in pixels. compute the min and max x/y values (in pixels) of all geoms in a panel.
  - when axis zoom is specified, the min/max values should be the same before and after a click.
  - when no axis zoom is specified, the min/max values should be different before and after a click.

faizan-khan-iit:
@tdhock I added the tests as discussed. I included the basic example I used while testing. I also included a test for multiple single selectors in a different file [here](https://github.com/tdhock/animint/pull/158/commits/88b3b7396f5a9667c6cda98c54c0e60d988b2df3). Currently it will just fail with an error.

Also I determined why the widerect test is failing. The tallrect in PANEL 4 is not rendering properly because the scale y ranges are reversed [here](https://github.com/tdhock/animint/blob/687ca04aefc4d898440f79a48a46aea4179c11df/inst/htmljs/animint.js#L1427).  Hence the height is negative. It is probably because of the re-scaling of data earlier. I will look into it a little more.

Any comments on the progress yet?

faizan-khan-iit:
I was also thinking of adding some more error checking to provide the user with informative warnings/messages in the code. Any particular ideas?
Currently I just provide a [single basic warning](https://github.com/tdhock/animint/pull/158/commits/7434908277395879457f919e711871b2f85cd16a#diff-fddb8085c2de24f7941a89d932954d47R1767) if no `subset_domain` is found.

R/animint.R
```diff
@@ -1633,7 +1638,138 @@ animint2dir <- function(plot.list, out.dir = tempfile(),
...
-
+  
+  ## Compute domains of different subsets, to be used by update_scales
+  ## in the renderer
+  compute_domains <- function(built_data, axes, geom_name,
+                             var, split_by_panel){
+    # Different geoms will use diff columns to calculate domains for
+    # showSelected subsets. Eg. geom_bar will use 'xmin', 'xmax', 'ymin',
+    # 'ymax' etc. while geom_point will use 'x', 'y'
+    domain_cols <- list(bar=c(paste0(axes, "min"), paste0(axes, "max")),
```
tdhock: this is fine but make sure to stop with an error if there is some geom which is not recognized

tdhock:
I would say that for loops are fine, don't worry about replacing them with apply functions. 

glad to see you have a test with several single selection variables. I think it is important that it works for that case.

not sure I understand what you mean by error checking. do you have an example/test where that warning appears about no subset_domain found?

faizan-khan-iit:
@tdhock Regarding the test for more than one single selection variables, I guess we have to compute every possible combination and use that for domains, ticks and grid updates. For now I will try and implement that.

As for the subset_domain warning, are you referring [this](https://github.com/tdhock/animint/pull/158/commits/c904489dd1f0f91ad57cf1ca03c49c84d406362a)?
It happens when the user has not specified the `selector.types` as `single`.

We will have to edit it a little once we get the test for more than one single selection variable.

faizan-khan-iit:
@tdhock I implemented the multiple showSelected vars updates.

Progress: (http://bl.ocks.org/faizan-khan-iit/raw/56afc5e59ba8a24b1ad0c8c6257c7dfe/)
Please note the cases when the subset has no data or a single unique data value. Any comments on that?

I will write some tests after I test the code on some other examples.

tdhock:
Hey Faizan this looks like a step in the right direction, good job!

Good point about when there is no data, and when there is only 1 data point. In both cases the axes are undefined. In that case I suggest warning the user.

faizan-khan-iit:
@tdhock I tested the code on a couple more examples, and it seems to be working fine. I will try to construct an example with 3 showSelected vars (2 showSelected vars seem good for now). Maybe with some dummy data like [this one](http://bl.ocks.org/faizan-khan-iit/raw/7f3f909d4800c77f5d2a8f5d6ff5f2b6/). I just added a random column for subsetting data to the example that I was using. Once I'm done I will write some tests for this. 

I noticed a minor point. When there is only a single point, I return a range equal to the point value. But if the value is 0 itself, it returns zero for both max and min [here](https://github.com/tdhock/animint/pull/158/commits/4f2ede44f8627c09abf6afa7aad04f88d9ce3439#diff-fddb8085c2de24f7941a89d932954d47R1720). You could see this by selecting `Rural` and changing the `District` in the above example. I could easily handle this case by returning a range [-1, 1] whenever we encounter this. But just asking in case you have any better ideas to return something else as a whole when there is a single data point.

tdhock:
my feeling is that if there really is a single data point then the user should not be doing axis updates, and we shouold warn them about that

but [-1, 1] is fine

faizan-khan-iit:
@tdhock Like I mentioned, I would like to test the code on some more geoms. We already have `point`, `bar` and `path`. If you have some examples of other geoms, I would be happy to use them! It would be better to have all the supported geoms work with this.

faizan-khan-iit:
@tdhock https://travis-ci.org/tdhock/animint/jobs/144237349#L2722
The widerect test has been failing since the PR opened, so we could assume the problem lies somewhere in the initial commits that I made. That's why I think the problem is that we do not rescale the data anymore. I will look into this more once the tests are done...

Also one more thing, when the plot renders initially, the axes do not zoom in automatically. This only happens once the user starts interacting with the plot. Would it be better that we automatically zoom in as soon as the plot loads?

And do we have to support selectors with `multiple` selection or not?

tests/testthat/test-renderer4-geoms-zoom.R
```diff
@@ -0,0 +1,97 @@
...
+no_updates_nodes1 <-
+  getNodeSet(info$html_updated1, '//g[@class="geom1_point_g"]//circle')
+no_updates_nodes2 <-
+  getNodeSet(info$html_updated2, '//g[@class="geom1_point_g"]//circle')
+no_updates_attrs1 <- sapply(no_updates_nodes1, xmlAttrs)[c("cx", "cy"), ]
+no_updates_attrs2 <- sapply(no_updates_nodes2, xmlAttrs)[c("cx", "cy"), ]
+no_updates_xranges1 <- range(as.numeric(no_updates_attrs1[1, ]), na.rm = T)
+no_updates_xranges2 <- range(as.numeric(no_updates_attrs2[1, ]), na.rm = T)
+no_updates_yranges1 <- range(as.numeric(no_updates_attrs1[2, ]), na.rm = T)
+no_updates_yranges2 <- range(as.numeric(no_updates_attrs2[2, ]), na.rm = T)
```
tdhock: hey @faizan-khan-iit can you please clarify this test by removing repetition? for example you could make a helper function `get_pixel_ranges` that takes the geom class name as argument...

```
get_pixel_ranges("geom2_point_g1")
get_pixel_ranges("geom3_point_g2")
```

tests/testthat/test-renderer4-geoms-zoom.R
```diff
@@ -0,0 +1,97 @@
...
+                                  data = mtcars)
+update_x <- no_updates+
+  theme_animint(update_axes=c("x"))
+update_y <- no_updates+
+  theme_animint(update_axes=c("y"))
+update_xy <- no_updates+
+  theme_animint(update_axes=c("x","y"))
+
+viz <- (list(g=no_updates, 
+             g1=update_x,
```
tdhock: instead of g, g1, g2, can you please change to more informative names? e.g. `neither` `x` `y` `both`

tests/testthat/test-renderer4-geoms-zoom.R
```diff
@@ -0,0 +1,97 @@
...
+xy_updates_xranges1 <- range(as.numeric(xy_updates_attrs1[1, ]), na.rm = T)
+xy_updates_xranges2 <- range(as.numeric(xy_updates_attrs2[1, ]), na.rm = T)
+xy_updates_yranges1 <- range(as.numeric(xy_updates_attrs1[2, ]), na.rm = T)
+xy_updates_yranges2 <- range(as.numeric(xy_updates_attrs2[2, ]), na.rm = T)
+
+
+test_that("geoms get zoomed-in upon changing selection", {
+  # no_updates
+  expect_false(isTRUE(all.equal(no_updates_xranges2, no_updates_xranges1)))
+  expect_false(isTRUE(all.equal(no_updates_yranges2, no_updates_yranges1)))
```
tdhock: to clarify your intent, could you please define `expect_not_equal` in helper-functions.R?

tests/testthat/test-renderer4-ticks-grids-updates.R
```diff
@@ -0,0 +1,192 @@
...
+  theme_animint(update_axes=c("x"))
+update_y <- no_updates+
+  theme_animint(update_axes=c("y"))
+update_xy <- no_updates+
+  theme_animint(update_axes=c("x","y"))
+
+viz <- (list(g=no_updates, 
+            g1=update_x, 
+            g2=update_y, 
+            g3=update_xy))
```
tdhock: same here, more informative names would be nice. any reason that these tests are in a separate file? If the viz list is exactly the same I would suggest combining these two test files.

tdhock:
no let's not support multiple selection for now

I looked at the test-widerect.R file and indeed the geom_tallrect is not rendering properly

I investigated by putting a breakpoint in the JavaScript code in the eActions function for geom_tallrect

screenshot-breakpoint

Then I printed out the values of the domain and range for both x and y scales

screenshot-scales-x-y-values

The domain values indicate that the scales are for the upper left panel. But the tallrect is rendered in the lower right panel. That is the source of the rendering error.

Please make sure to set `scales.x` and `scales.y` to the values for the correct PANEL.

DESCRIPTION
```diff
@@ -31,6 +31,6 @@ Suggests:
     httr,
     maps
 Remotes: 
-    ropensci/RSelenium,
-    hadley/ggplot2#1649
+    ropensci/RSelenium@22f06b9f2a675015d0daa3318319044e4e60d2fa,
+    faizan-khan-iit/ggplot2@5fb99d0cece13239bbbc09c6b8a7da7f86ac58e2
```
tdhock: By including the SHA1, the new version of devtools can skip installation, if the correct version is already installed

tdhock:
i'm trying to get this example to work with axis updates https://github.com/tdhock/animint-examples/blob/master/examples/chip.seq.R

to get those examples to work we need to make sure to include support for
- [ ] geom_segment(aes(x, y, xend, yend))
- [ ] geom_text(aes(x,y))
- [ ] geom_rect(aes(xmin, xmax, ymin, ymax))
- [ ] geom_point(aes(x,y))
- [ ] geom_ribbon(aes(x, ymin, ymax))
- [ ] geom_tallrect(aes(xmin, xmax))

faizan-khan-iit:
@tdhock Thanks for the tip.

I spent quite some time trying to get the `tallrect` and `widerect` to render properly and I could not find any problem with the `PANEL` that was being used for rendering. I observed that the problem was because earlier we reversed the range of the *real* y axis and the domain of the *fake* y axis ([see here](https://github.com/tdhock/animint/pull/158/files#diff-bfbd034b190fd160d92db9a24ea2a3c7L586)). So I did the same: https://github.com/tdhock/animint/pull/158/commits/91cd20205bb8c5d20b226e7ba044577cdf15f990

I also rewrote the tests a bit and changed the names accordingly.

tdhock:
If I remember correctly, we introduced the fake axes to deal with ggplot2's [0,1] built data coordinate system. Now that you are not using those [0,1] coordinates anymore, do you think we could delete the fake axis? (just use the real axis) It would simplify the code.

faizan-khan-iit:
@tdhock I already deleted the fake axes https://github.com/tdhock/animint/pull/158/files#diff-bfbd034b190fd160d92db9a24ea2a3c7L574

Now we use only one set of axes.

faizan-khan-iit:
What I meant was that I reversed the range instead of the domain

tdhock:
by the way sorry it took so long to figure out the fake axis problem!

I have done code review and I think this is really excellent work.

the only other thing I would like before merging with master is a few tests for geoms other than point (segment, text, ribbon, rect, tallrect, widerect). do you think that is easy/possible?

faizan-khan-iit:
@tdhock Yeah no problem! I will work on that next. I do not think it should get too complicated.

faizan-khan-iit:
@tdhock I have got the tests for `rect`, `text`, `ribbon` and `segment`. Basically I am checking for no warnings, axis ticks updates and grid updates for all of these.

As for `tallrect` and `widerect`, could you give me an example of how these would work with axes updates? Currently it is something like this: http://bl.ocks.org/faizan-khan-iit/raw/0574dd1f6c89097b5004e74b8ab9f21d/

tdhock:
wow, that example is really nice faizan

I would like to get this example to work with your code https://github.com/tdhock/animint-examples/blob/master/examples/chip.seq.R#L247 that example is rather complicated but hopefully should work with your code.I don't recommend adding this example as a test case (data too big and complicated).
There is a `geom_tallrect` with `showSelected=set.name` https://github.com/tdhock/animint-examples/blob/master/examples/chip.seq.R#L247 Its `xmin` and `xmax` values should be used in the x scale computation.

More generally, if `geom_rect` is already working with your code then I think it should be no problem to get tallrects working. `geom_rect` is used for both x and y scale computation, but tallrect should only be used for x scales (xmin/xmax) and widerect should only beused for y scales (ymin/ymax).

Does that make sense?

faizan-khan-iit:
@tdhock In that case it should not be too difficult. I will give it a try.

Also, would you like zooming in tests for the new geoms (`rect`, `segment` etc.). I have not added them yet cause they were a little different from `geom_point`. Or are the current tests with ticks and grid updates fine?

tdhock:
yes the current tests are fine I think, no need to test everything for all geoms

however when I run `animint2dir(unaligned.update)` from 

I get the following error and warnings, which I suppose will be fixed when you implement support for tallrect.

```
Error in min(sapply(all_vals, "[[", 1)) : 
  invalid 'type' (list) of argument
In addition: Warning messages:
1: In animint2dir(unaligned.update, "chip-seq-unaligned-update") :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=complexity.i), problem: geom14_text_error
2: In animint2dir(unaligned.update, "chip-seq-unaligned-update") :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=complexity.i), problem: geom15_text_error
3: In animint2dir(unaligned.update, "chip-seq-unaligned-update") :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=complexity.i), problem: geom16_text_error
4: In animint2dir(unaligned.update, "chip-seq-unaligned-update") :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=complexity.i), problem: geom17_path_roc
5: In animint2dir(unaligned.update, "chip-seq-unaligned-update") :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=complexity.i), problem: geom18_point_roc
6: In animint2dir(unaligned.update, "chip-seq-unaligned-update") :
  to ensure that smooth transitions are interpretable, aes(key) should be specifed for geoms with aes(showSelected=complexity.i), problem: geom21_ribbon_probability
7: axis updates currently do not work for geom_tallrect 
8: In animint2dir(unaligned.update, "chip-seq-unaligned-update") :
  axis updates only work for single selection variables
9: axis updates currently do not work for geom_tallrect 
```
the "axis updates do not work for geom_tallrect" warning is nice and informative

however I think the "axis updates only work for single selection variables" warning is confusing to me. I'm guessing that it came from the `geom_tallrect(aes(fill=annotation))` https://github.com/tdhock/animint-examples/blob/master/examples/chip.seq.R#L340 which creates a multiple selection variable for that interactive legend.

http://cbio.mines-paristech.fr/~thocking/figure-chip-seq-unaligned/index.html

However I don't think there is any reason to warn the user in that case. Instead, can you please just ignore the multiple selection variable for the purposes of scale computation? e.g. there are two tallrects shown for each value of `showSelected=set.name` so just compute the min/max over those two tallrects. And actually the tallrects should not affect the min/max in this particular example, because there are other geoms which have more extreme values (the blue geom_ribbon in the "learned difference function" plot, and the grey geom_rect in the "ChIP-seq signal pair" plot)

tdhock:
maybe add a simple test that mimics this example? e.g. `geom_tallrect(aes(fill=annotation, showSelected=set.name))` or something like that.

faizan-khan-iit:
@tdhock The warning is supposed to be given when after searching the whole plot, we still do not find a `subset_domain`, in which case we do not have any axis updates, and the viz works like normal. I think this is because the plot does not return any domain in this case (as we currently ignore `tallrect`). Would you like me to remove this? In that case if a user forgets to set selector vars to `single`, he will just get a normal viz with no updates, so I would like to keep this.

Regarding the error, I think it is quite unexpected that we run into one here. I will look into this and try to figure out what the problem is..

tdhock:
ok I understand your warnings now. 

please change them to clarify.

"axis updates have not yet been implemented for geomXX_point_plotName"

"update_axes specified for X axis on plot PLOTNAME but found no geoms with showSelected=singleSelectionVariable, so created a plot with no updates for X axis"

faizan-khan-iit:
@tdhock I was trying to get the `ChIP-seq` example to work and it somewhat works, but there are two problems:

1. The multiple selectors do not update themselves when the single selectors update for axis zooms. So we have to change the multiple selectors too for them to update. Once we manually update them, it works fine.
2. I got into a problem where the geoms have different single `showSelected` vars in diff geoms in the same plot. For example, in the `ChIP-seq` example we have other single selection `showSelected` vars except for `set.name` in `geom_text` and the two `geom_text`s. This is not handled correctly by the current code.

I have a basic idea of how to fix the second problem. Any ideas on the first one?

tdhock:
glad you have an idea about how to fix problem (2). can you think of a simple test case to write, that would fail on the current code?

About problem (1) I do not really understand. Which multiple selection variable is not updating? annotation? Could you please push your current code, and a link to a simpler example that captures the problem?

faizan-khan-iit:
@tdhock Could you please run the ChIP-seq example -`animint2dir(unaligned.update)` - again on your system and give your feedback regarding the updates of axes?

As for problem (2), I am a bit stuck here unfortunately.
I have a dummy example:
```
set.seed(123)
data_f1 <- data.frame(a=runif(30, 1, 30), b=sample(1:30))
data_f1$c <- data_f1$a + runif(30, 1, 10)
data_f1$d <- data_f1$b + runif(30, 1, 10)
data_f1$ss1 <- as.factor(1:3)
data_f1$ss2 <- as.factor(c("alpha", "beta"))
data_f1$ss3 <- as.factor(c("A", "A", "B"))

plot1 <- ggplot() + 
  geom_point(aes(a,b,
                 showSelected1=ss1,
                 showSelected2=ss2),
             data = data_f1) +
  geom_segment(aes(x=a,y=b,xend=c,yend=d,
                 showSelected3=ss3),
             data = data_f1) +
  theme_animint(update_axes=c("x"))
viz <- list(p=plot1)

viz$selector.types <- list(ss1="single", ss2="single", ss3="single")
animint2dir(viz, "./test-dir", open.browser = T)
```

In this example the axes updates do not work (they do not create an error/warning) because I cannot calculate the domains of all the combinations in the. For example, if I use `ss2` in `geom_segment` too, all works well, as I am able to calculate all the possible combinations from the `built_data`. But in this case, since I do not have all the appropriate columns in the `built_data` to calculate the combinations of `ss1`, `ss2` and `ss3` at a time, it messes up the renderer implementation, and no axes are updated.

So I wanted some help here. Are we supposed to support cases like these? I know this example does not make much sense, but we do have cases like these, right?

I tried calculating the ranges geom by geom for every selector in the whole plot, but it's no good. Any good ideas how to work around this?

Currently we will have axes updates as long as one of the supported geoms has all the singly selected vars specified in it.

faizan-khan-iit:
Also for some reason the `renderer-3` tests are not terminating and continuing to execute `renderer-4` tests in Travis, leading to an error - https://travis-ci.org/tdhock/animint/builds/153328070

tdhock:
that seems like a pretty contrived example, not sure if it will be a problem in real data sets in practice.

but maybe this result means that there is some inherent limitation to doing axes update computations in the compiler? maybe it would be better to one day (not in this PR) switch the computation to the renderer?

about the travis problem, it may be a problem on their end, let's just wait and see if they fix it tomorrow.

finally about the chip.seq data example, there are a couple of problems. one problem was in my design, I forgot to delete the geom_text labels in the last plot. I just pushed those changes.

but the main problem is that the tallrects in the "learned difference function" plot do not appear in the right place when the plot is first rendered. but then if I click the point in the first plot next to chr21, which selects `set.name=train` (which was already supposed to be selected when the plot is first rendered) then they do appear in the right place. the correct behavior can be seen here http://cbio.mines-paristech.fr/~thocking/figure-chip-seq-unaligned/ (with no axis updates) and the incorrect behavior rendered by the current code is http://cbio.mines-paristech.fr/~thocking/figure-chip-seq-unaligned-update/ any idea why that may be happening?

tdhock:
maybe we should just merge the PR in its current state, and worry about these more complicated examples later? If you agree then please add these examples that don't work as failing tests in another PR, and you can merge this one as it is. +1

by the way I am going on vacation with limited computer access for the next couple of weeks. I won't be able to do any programming or debugging but I should be able to comment on your PR. And I will definitely be passing you for GSOC! Thanks for all the great work Faizan.

faizan-khan-iit:
@tdhock Thanks a lot Toby! Really glad to hear that!

Regarding the computation limitation in the compiler, I agree too. Maybe it would be more practical to move this to the renderer. Hopefully we may also be able to support `multiple` selector.types!

Regarding the issue in the `ChIP-seq` example, I think it is because I do not call `update_scales` and related functions until the user clicks/changes any selection (as it is called by `update_selector` function https://github.com/tdhock/animint/blob/9723e9d2947f42fc56486488b4baa292699a8363/inst/htmljs/animint.js#L1921). I should probably call it once more after the plots have been drawn initially if there is any plot with `update_axes` specified. Is it working fine except for this issue?

I will work on this till Sunday, and try to sort out the issues with our current implementation. After that I will merge this branch as you said.

Also would you like some changes in the documentation for the `update_axes` feature?

tdhock:
yeah the problem with the chip.seq data viz is probably that you are not calling update_scales initially. you should definitely do that, probably after add_selector and add_plot https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L1959

and in fact I think there is another problem with the chip.seq data viz -- there are no grey `geom_rect` elements drawn at all in the bottom plot.

about the docs, go ahead and change any Rd files, and please add a section to this wiki page https://github.com/tdhock/animint/wiki/Advanced-features-present-animint-but-not-in-ggplot2

tests/testthat/test-renderer4-update-axes-multiple-ss.R
```diff
@@ -0,0 +1,312 @@
...
+# Plots with axis updates
+no_updates <- plot3
+
+update_x <- no_updates+
+  theme_animint(update_axes=c("x"))
+update_y <- no_updates+
+  theme_animint(update_axes=c("y"))
+update_xy <- no_updates+
+  theme_animint(update_axes=c("x","y"))
+
```
cpsievert: Is there any reason why we couldn't make the default to update both axes?

faizan-khan-iit: @cpsievert Basically if we pass `update_axes` in `theme_animint`, are you suggesting we should have axes updates for both `x` and `y` as a default, until a user explicitly specifies an axes? That is certainly a good point. In most cases we would want axis updates for both axes.

cpsievert: exactly

faizan-khan-iit: @cpsievert So we could have something like 
```
update_xy <- no_updates+
  theme_animint(update_axes=TRUE)
```
that should update both axes in a plot. It looks nice to me. Should I edit the compiler for this?

cpsievert: No, I mean we should enable that behavior by default. Most people don't take the time to read documentation on our wiki page ;)

You could slip in a line like this here -- https://github.com/tdhock/animint/blob/29e342c324c63fed083bb53c177111cc7417916f/R/theme.R#L11

```r
if (is.null(elements[["update_axes"]])) {
  elements[["update_axes"]] <- c("x", "y")
}
```

faizan-khan-iit: @cpsievert That is certainly true. Most users tend to skip the documentation.

But I am a little hesitant in doing this as we do not support `multiple` selection vars as of now.  So users not specifying their selection vars as `single` will keep getting warnings they do not expect. This could get a bit annoying when they get warnings of the likes of "created plot with no updates" and stuff, don't you think? So I would suggest the `update_axes=TRUE` approach for now...

Maybe once we have implemented this for both types of selections and sorted out the current bugs, we could definitely do that. Any thoughts?

cpsievert: I believe single selection is the default, so I'd be OK with throwing an informative warning/message if the user opts into multiple (and hasn't specified `update_axes`). What do you think @tdhock?

faizan-khan-iit: I think we have multiple selection as default, right? https://github.com/tdhock/animint/blob/3358aa3634cfcac7a7cc3e4b8ee5ec1620985045/R/animint.R#L103-L108

Should we switch to single selection?

tdhock: I'm sorry but I definitely do NOT want update_axes by default.

For me in the vast majority of cases I want to see the data on a common axis, to facilitate comparisons between selection states.

faizan-khan-iit:
@tdhock Regarding the initial updates issue, I have implemented it. Now we have axis updates as soon as the plot loads with `first` selections.
As for the `geom_rect` problem, it should be fine now. It was a minor mistake on my part. But since the last plot has `sample1` and `sample2` as showSelected in different geoms, it will still be a bit of a problem. Specifying these as `multiple` selection variables seems to work well..

I have also updated the tests so that the tests fail properly when there are no axes updates when we expect.

Should I merge this with `master` after updating the `DSCRIPTION` and `NEWS`? I will open another PR like we discussed.

faizan-khan-iit:
Looks like Travis has a problem when the `tooltip-interactivity` test executes after other tests.
@tdhock Any particular reason why the test was moved to `renderer3` instead of `renderer4`?

faizan-khan-iit:
@tdhock Looks like it's wercker's turn to fail now that I moved the test to `renderer4`. For now I will move the test to `renderer5`. Hopefully we can come up with a solution in the future.

I will go ahead and merge this now.

PR #160, opened 2016-08-04
New feature legend decode added

viswaraavi:
When the user hovers on the element the respective legend text gets highlighted and gets bold with increase in font size. When hovers out the legend text becomes normal.

### DIFF
--- a/inst/htmljs/animint.js
+++ b/inst/htmljs/animint.js
@@ -42,7 +42,7 @@ var animint = function (to_select, json_file) {
   // replacing periods in variable with an underscore this makes sure
   // that selector doesn't confuse . in name with css selectors
   function safe_name(unsafe_name){
-    return unsafe_name.replace(/[ .]/g, '_');
+    return unsafe_name.replace(/[^-A-Za-z0-9_]/g,'');
   }
   function legend_class_name(selector_name){
     return safe_name(selector_name) + "_variable";
@@ -1229,7 +1229,9 @@ var animint = function (to_select, json_file) {
 
         // Need to store the clickSelects value that will
         // be passed to the selector when we click on this
-        // item.
+        // item And also showSelectedlegendcolour is added for
+        //legend decoding
+        d.showSelectedlegendcolour=keyed_data[d.value][0].showSelectedlegendcolour;
         d.clickSelects = keyed_data[d.value][0].clickSelects;
         return d;
       });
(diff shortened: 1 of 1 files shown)


viswaraavi:
I am still working on this to make the build pass. I think this change "return unsafe_name.replace(/[^-A-Za-z0-9_]/g,'');" is what making the tests to break otherwise I could not see any reasons why tests are breaking

tdhock:
Hey @viswaraavi thanks for the PR. Do you have any examples of this feature working? If so, can you please run `animint2gist` and post a link to the resulting data viz?

After that, can you please think about writing a test for this new feature? Apparently we can simulate mouseover/out events in RSelenium https://github.com/ropensci/RSelenium/issues/59

viswaraavi:
This is the example for the new feature added
http://bl.ocks.org/viswaraavi/raw/2e323641b70f9ca904dd44e11b6bc0e9/

tdhock:
wow very nice work!

in my opinion it would be less visually startling if the text size did not change. that has the effect of moving the entire legend. it would be better if the only thing that changes is the legend entry itself. can you please try just using bold text, without changing the text size?

also have you considered what to do for multiple legends? you could very well have a single data point which should highlight an entry in more than one legend. for example `aes(color=sex, fill=smoker)`

viswaraavi:
Now i think  the support for multiple legends work 
http://bl.ocks.org/viswaraavi/raw/e85bba5637ebd43e237dc8a8a6e7ad2e/
But still i need to know in what other ways can the multiple legend will be created

viswaraavi:
1.I could not understand why animint2gist tests are failing.
2.In case of world bank dataset the legend does not highlight for population, this is because the data which is used to draw the geom_point in d3 does not have showselectedlegendsize as attribute instead has size as attribute. Can i know the reason for this?. If this is the case then somehow I should map between the size of the data and legend  or else i have to change the compiler code to make this work

tdhock:
looks good for the example with multiple legends. thanks for not changing the text size.

however the entire legend is still moving. one way to fix that would be to increase the height of the legend SVGs from 14 to 16 https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L1876

animint2gist tests always fail on PRs from other people, so that is fine, don't worry about it

about the size legend in the worldbank example - it is a continuous legend, not a discrete legend. showSelected variables are automatically created only for discrete legends. for now (this PR) I think it is OK to ignore continuous legends.

for a future PR it would be nice to do something like http://www.highcharts.com/maps/demo/map-drilldown/dark-unica for continuous legends

viswaraavi:
I will try to fix the problem with the height. And try to write the tests for mouse hover events

tdhock:
any update @viswaraavi ?

viswaraavi:
l will start working on it again. May be this weekend. I was having classes

tdhock:
hey @viswaraavi I merged your other PR, so it looks like this one now has some conflicts that need to be resolved.

By the way, I added you as a collaborator, so now you have the power to add whatever you like to animint. However, please do not push changes to master, unless absolutely necessary. For future PRs (not this one), please use branches of tdhock/animint (then the animint2gist test will not fail).

Issue #162, opened 2016-08-05
Path/line grouping


cpsievert:
It appears the group aesthetic is ignored in this plot (lines are connecting values for different cities, but they shouldn't)

```r
library(animint)
p <- ggplot(txhousing) +
  geom_line(aes(x = date, y = median, group = city), alpha = 0.3)
animint2dir(list(p = p))
```

Also, if I add `showSelected = city`, no lines render:

```r
p <- ggplot(txhousing) +
  geom_line(aes(x = date, y = median, group = city, showSelected = city), alpha = 0.3)
animint2dir(list(p = p))
```

cpsievert:
```r
Session info ------------------------------------------------------------
 setting  value                       
 version  R version 3.3.1 (2016-06-21)
 system   x86_64, darwin13.4.0        
 ui       RStudio (0.99.1266)         
 language (EN)                        
 collate  en_US.UTF-8                 
 tz       America/Chicago             
 date     2016-08-05                  

Packages ----------------------------------------------------------------
 package     * version    date       source                            
 animint     * 2016.07.12 2016-07-21 Github (tdhock/animint@3a11efd)   
 assertthat    0.1        2013-12-06 CRAN (R 3.3.0)                    
 colorout      1.1-2      2016-05-10 Github (jalvesaq/colorout@6538970)
 colorspace    1.2-6      2015-03-11 CRAN (R 3.3.0)                    
 curl          1.1        2016-07-26 CRAN (R 3.3.0)                    
 DBI           0.4-1      2016-05-08 CRAN (R 3.3.0)                    
 devtools    * 1.12.0     2016-06-24 CRAN (R 3.3.0)                    
 digest        0.6.10     2016-08-02 cran (@0.6.10)                    
 dplyr       * 0.5.0      2016-06-24 CRAN (R 3.3.0)                    
 fortunes      1.5-3      2016-05-20 CRAN (R 3.3.0)                    
 ggplot2     * 2.1.0.9000 2016-08-05 Github (hadley/ggplot2@84c203a)   
 git2r         0.15.0     2016-05-11 CRAN (R 3.3.0)                    
 gtable        0.2.0      2016-02-26 CRAN (R 3.3.0)                    
 htmltools     0.3.5      2016-03-21 CRAN (R 3.3.0)                    
 htmlwidgets   0.6        2016-02-25 CRAN (R 3.3.0)                    
 httpuv        1.3.3      2015-08-04 CRAN (R 3.3.0)                    
[...]

PR #166, opened 2016-08-13
NA separate lines test fails

tdhock:
when there are missing values in a visual characteristic for a line (x,y,color,linetype,size) ggplot2 splits the line into separate lines

also there seems to be a problem with the common chunk computation for the txhousing test data set


tests/testthat/helper-functions.R
```diff
@@ -199,12 +199,14 @@ getTextValue <- function(tick)xmlValue(getNodeSet(tick, "text")[[1]])
...
-    attr.mat["style", ]
-  }else{
-    rep("", length(node.list))
-  }
+  style.vec <- sapply(node.list, function(node){
+    attr.vec <- xmlAttrs(node)
+    if("style" %in% names(attr.vec)){
+      attr.vec[["style"]]
+    }else{
+      NA
```
tdhock: this fixes a bug that was returning a whole vector of NA when only one node was missing the style attribute

tdhock:
hey @cpsievert this PR fixes your issue #162 so could you please code review and +1 before I merge with master?

R/animint.R
```diff
@@ -966,11 +960,32 @@ saveLayer <- function(l, d, meta){
...
+    is.missing <- apply(is.na(g.data), 1, any)
+    diff.vec <- diff(is.missing)
+    new.group.vec <- c(FALSE, diff.vec == 1)
+    for(chunk.col in sp.cols){
+      one.col <- g.data[[chunk.col]]
+      is.diff <- c(FALSE, one.col[-1] != one.col[-length(one.col)])
+      new.group.vec[is.diff] <- TRUE
+    }
+    subgroup.vec <- cumsum(new.group.vec)
+    g.data$group <- subgroup.vec
```
tdhock: cumsum is a pretty fast way to get a unique subgroup id -- this replaces group when there is for example a line with NA which should be split into several lines

R/animint.R
```diff
@@ -966,11 +960,32 @@ saveLayer <- function(l, d, meta){
...
+  }
+
   ## Determine if there are any "common" data that can be saved
   ## separately to reduce disk usage.
   data.or.null <- getCommonChunk(g.data, chunk.cols, g$aes)
   g.data.varied <- if(is.null(data.or.null)){
-    split.x(g.data, chunk.cols)
+    split.x(na.omit(g.data), chunk.cols)
```
tdhock: there should not be any missing data in tsv files

R/animint.R
```diff
@@ -1022,69 +1037,89 @@ getCommonChunk <- function(built, chunk.vars, aes.list){
       built[, col.name] <- paste(built[, col.name])
     }
   }
-  built.by.chunk <- split(built, built[, chunk.vars], drop = TRUE)
```
tdhock: i removed this split by all chunk variables, in order to speed up the common chunk computation

R/animint.R
```diff
@@ -1022,69 +1037,89 @@ getCommonChunk <- function(built, chunk.vars, aes.list){
...
+
+  checkCommon <- function(col.name){
+    for(group.name in names(built.by.group)){
+      data.vec <- built.by.group[[group.name]][[col.name]]
+      if(group.size <- each.group.same.size[[group.name]]){
+        not.same.value <- data.vec != data.vec[1:group.size]
+        if(any(not.same.value, na.rm=TRUE)){
+          ## if any data values are different, then this is not a
+          ## common column.
+          return(FALSE)
```
tdhock: as soon as we have figured out that `col.name` is not a common column, we can exit immediately

R/animint.R
```diff
@@ -1022,69 +1037,89 @@ getCommonChunk <- function(built, chunk.vars, aes.list){
...
-    for(col.name in col.name.vec){
-      value.list <- lapply(values.by.chunk, function(df)df[[col.name]])
-      is.common.mat[group.name, col.name] <- if(same.size.chunks){
-        value.mat <- do.call(cbind, value.list)
-        least.missing <- which.min(colSums(is.na(value.mat)))
-        value.vec <- value.mat[, least.missing]
-        one.group.info[[col.name]] <- value.vec
-        all(value.vec == value.mat)
+
+  built.by.group <- split(built, built$group)
```
tdhock: this re-design of the common chunk computation is faster since it only has 1 level of for loops (through groups), as opposed to the previous code which had two levels (through groups and chunks)

tests/testthat/test-renderer1-PeakConsistency.R
```diff
@@ -124,17 +124,17 @@ test_that("15 segments of both colors in last plot", {
...
+## test_that("showSelectedlegendcolour is truth", {
+##   tsv.path <-
+##     file.path("animint-htmltest", "geom6_vline_signals_chunk_common.tsv")
+##   common.df <- read.table(tsv.path, comment.char="", header=TRUE)
+##   computed.vec <- paste(common.df$showSelectedlegendcolour)
+##   expected.vec <- rep("truth", length(computed.vec))
+##   expect_identical(computed.vec, expected.vec)
+##   tsv.path <-
+##     file.path("animint-htmltest", "geom6_vline_signals_chunk1.tsv")
+##   varied.df <- read.table(tsv.path, comment.char="", header=TRUE)
```
tdhock: I turned off this test because it was failing, and not a big deal. I think it was because the varied data files only had the group column (and no other data). for now I think it is OK to ignore, but I left it as comments in case we want to address that one day.

tests/testthat/test-renderer1-text.R
```diff
@@ -1,7 +1,16 @@
...
 data(WorldBank, package = "animint")
-wb  <- WorldBank[WorldBank$year == 2010,]
+wb2010 <- subset(WorldBank, year==2010)
+subset(wb2010, population==min(population))
+### This test does not pass if we use the wb2010 data set above rather
+### than the wb data set below. The problem is that the scales are
+### trained before NAs are removed from the data. There are NA values
+### in life.expectancy/fertility.rate for Tuvalu, but not in
+### population. So Tuvalu will not be rendered on the plot, and in
+### fact there will be no text element with fontsize=10!
```
tdhock: did you know that scales are trained before removing NA? I did not! It results in some weird scales sometimes https://github.com/hadley/ggplot2/issues/1711

tests/testthat/test-renderer1-tooltip.R
```diff
@@ -107,22 +108,13 @@ info <- animint2HTML(viz)
 
 test_that("tooltip works with href",{
   # Test for bug when points are not rendered with both href + tooltip
-  
-  # See that points are rendered for every country. This test may need to be
-  # changed soon!!! See the viz on the link below:
-  # http://bl.ocks.org/faizan-khan-iit/449b84b2e6b55ca5e39ea9aaad97e27e
-  # You may notice a point on the very top-left of the viz, outside the plot.
-  # These are the points that have NaN as their value in the DOM, arriving from
-  # NA values in the tsv. Should be an easy fix. But leaving a comment here
```
tdhock: @faizan-khan-iit this PR fixes that, by removing the NA data points from the tsv files.

tests/testthat/test-renderer4-NA-separate-lines.R
```diff
@@ -0,0 +1,41 @@
...
+  path.list <- getNodeSet(info$html, xpath)
+  opacity.str <- getStyleValue(info$html, xpath, "opacity")
+  opacity.num <- as.numeric(opacity.str)
+  hilite.list <- path.list[opacity.num == 0.6]
+  expect_equal(length(hilite.list), 3)
+})
+
+test_that("three <path> rendered for selected San Marcos", {
+  path.list <- getNodeSet(info$html, '//g[@class="geom2_line_selected"]//path')
+  expect_equal(length(path.list), 3)
```
tdhock: @cpsievert anything else you wanted to test in this data viz?

tests/testthat/test-renderer1-tooltip.R
faizan-khan-iit: @tdhock As I was testing the axis updates, [with this WorldBank example](http://bl.ocks.org/faizan-khan-iit/raw/a3e19ad00691d2dafffca97e9e4b0576/), I actually got seemingly wrong range values for zooming of geoms. E.g. upon selecting North America, the geoms do not cover the whole PANEL area... Removing NAs from the original data seemed to work fine.
Basically we remove the NAs before saving to tsv now, is that correct? In that case this example should also work fine..

tdhock: yes this PR should solve your problem. but for you may just want to remove NA from the data set before passing it to your geom.

tests/testthat/test-renderer4-NA-separate-lines.R
```diff
@@ -0,0 +1,41 @@
+acontext("NA separate lines")
+
+data(txhousing)
+
+san.marcos <- subset(txhousing, city=="San Marcos")
+## ggplot2 draws separate lines when there are missing values.
+ggplot()+
+  geom_line(aes(x = date, y = median),
+            data=san.marcos)
+
```
cpsievert: Perhaps we should comment out this code?

tests/testthat/test-renderer4-NA-separate-lines.R
cpsievert: Looks good to me! 👍

PR #169, opened 2016-08-23
clean up code

### LINKED ISSUE #168: Link broken in README
The link to long table of related work in the `README.md` file seems to be broken. The link points to this location: https://github.com/tdhock/animint/blob/master/etc/references.org

Ideally it should point to this one: https://github.com/tdhock/animint/blob/master/inst/etc/references.org

faizan-khan-iit:
This PR is to clean up the code that may be redundant now. e.g. https://github.com/tdhock/animint/blob/master/R/animint.R#L25-L40
or https://github.com/tdhock/animint/blob/master/inst/htmljs/animint.js#L1421-L1431
Also fixes #168

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -2,7 +2,7 @@ Package: animint
 Maintainer: Toby Dylan Hocking <tdhock5@gmail.com>
 Author: Toby Dylan Hocking, Susan VanderPlas, Carson Sievert, Kevin Ferris, Tony
     Tsai, Faizan Khan
-Version: 2016.08.21
+Version: 2016.11.02
 License: GPL-3
 Title: Interactive animations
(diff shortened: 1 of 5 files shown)


faizan-khan-iit:
Why are we doing this twice?
https://github.com/tdhock/animint/blob/master/R/animint.R#L42-L47
https://github.com/tdhock/animint/blob/master/R/animint.R#L142-L147

I think the second call is redundant... Could someone confirm?

tdhock:
thanks for the cleanup

yes it seems redundant and doesnt break any of the tests so that looks good to me +1

faizan-khan-iit:
@tdhock @cpsievert Anything else that could use some review? If not we could close this..

cpsievert:
👍

PR #171, opened 2016-11-02
Animint fun

tdhock:
rather than explicitly writing `structure(list(...), class="animint")` we can write `animint(...)`


inst/examples/test_knit_print.Rmd
```diff
@@ -18,9 +18,9 @@ knit_meta()
 library(ggplot2) 
 library(animint)
 dat <- data.frame(x = 1:10, y = 1:10, label = rep(c("a178", "b934"), 5))
-q2 <- list(q = qplot(x, y, data = dat, colour = label, 
-                     xlab = "first plot with color legend"))
-structure(q2, class = "animint")
+animint(q = qplot(
+  x, y, data = dat, colour = label, 
+  xlab = "first plot with color legend"))
```
tdhock: here I add a test for the new functionality in the existing test for rendering Rmd

tdhock:
hey @cpsievert any comments about this new functionality? I remember during your GSOC you said that you would rather be explicit and use `structure` to attach a class to an animint viz list. However, during my interactive graphics tutorial at useR2016, one of the people suggested to have a function like this.

